PR #1, opened 2020-11-01
worked design proposal for supervised forecasting

fkiraly:
worked design proposal for supervised forecasting

@partmor, @mloning

### DIFF
--- a/steps/04_supervised_forecasting_api/step.md
+++ b/steps/04_supervised_forecasting_api/step.md
@@ -70,22 +70,96 @@ There is a temporal continuity and association assumed between "past" and "futur
 
 The template class `BaseSupForecaster` inherits from `sklearn` `BaseEstimator`.
 
+It has the following methods:
 
+`fit(X : Panel, fh=None : ForecastingHorizon, cutoffs=None : np.array) -> self`  
+Behaviour: fits the model to training data `X`  
+Args:  
+`X` - `Panel` type, independent instances of training data
+`fh` - forecasting horizon object, optional  
+`cutoffs` - same length as `X`, i-th entry is cut-off time points (past vs future) of the i-th sample in `X`
+
+`fit_forecast(X : Panel, fh=None : ForecastingHorizon) -> self`
+Behaviour: fits the model to past of prediction data `X` - usually requires `fit`  
+Args:  
+`X` - `Panel`, independent instances of prediction data  
+`fh` - forecasting horizon object, optional - should only be passed if not passed in `fit`  
+
+`predict(X=None : Panel, fh=None : ForecastingHorizon, return_pred_int=False, alpha=DEFAULT_ALPHA)-> Panel`  
+Behaviour: makes prediction for prediction series in `X` - requires `fit`, but `fit_forecast` is optional; if requires `fit_forecast` but not run, runs it first  
+Args:  
+`X` - `Panel`, independent instances of prediction data  
+`fh` - forecasting horizon object, required if not passed earlier  
+`return_pred_int` - whether prediction intervals should be returned  
+`alpha` - prediction interval alpha  
+Returns:  
+`Panel` type object with predictions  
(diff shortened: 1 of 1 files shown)


steps/04_supervised_forecasting_api/step.md
```diff
@@ -1,15 +1,165 @@
...
 First models to implement:
+
 * reduction to supervised tabular
 * mixed effects models
 * reduction 2-step: use only constant variables for mean then another estimator for forecast
 
 Other resources
+
 * Master thesis (to be uploaded)
 * previous design notes (will see if I can find any)
```
mloning, line 165: We can remove this, you mention all of these points in more detail above
```suggestion
```

mloning (commented):
Looks good to me, happy to merge this with one minor comment

mloning:
One question: what's the reason for using fit(X, cutoffs=None, ...) over fit(X, Y=None, ...)?

fkiraly:
>One question: what's the reason for using fit(X, cutoffs=None, ...) over fit(X, Y=None, ...)?

What would `Y` stand for, here?
If you suggest a simple rename, then I would oppose this on the grounds of user expectation: `Y` is the thing to be predicted, usually - the cut-offs are not the thing to be predicted though.

steps/04_supervised_forecasting_api/step.md
fkiraly: sure - just thought to keep it for completeness

fkiraly:
>I will play around with some Python code when I find some time to try out different things.

You also need to leave some of the playing to @partmor :-)

steps/04_supervised_forecasting_api/step.md
mloning: But they will be in the git history, no?

mloning:
I was thinking of `Y` being the actual data for the "target" time window, so coming from something like this:
```python
X, Y = data[:, :, :cutoff], data[:, :, cutoff:]
```
This splits the data outside of the estimator rather than inside as in your proposal. I'm just wondering what the pros and cons are for each?

mloning:
One other suggestion, I've come to prefer the term panel forecasting over supervised forecasting, panel forecasting seems to be widely used already, aligns with our panel as the name for the data format, and is a bit shorter too.

fkiraly:
>This splits the data outside of the estimator rather than inside as in your proposal. I'm just wondering what the pros and cons are for each?

My reason for that was:
* in general there does not need to be a designated cut-off - for example, if all the input series are samples from a (panel) ARMA process (or considered as such).
* the "y" predicted in prediction mode, specified by `fh`, does in general not need to agree with the "y" in the training set.

The cut-off time or index may also be different for the observed series - not sure whether we want to cover that case though.

fkiraly:
>One other suggestion, I've come to prefer the term panel forecasting over supervised forecasting, panel forecasting seems to be widely used already, aligns with our panel as the name for the data format, and is a bit shorter too.

I think the two are different tasks!

In panel forecasting, to the best of my knowledge, you try to forecast the future of the entire panel. While you assume that there are i.i.d. instances, there is no "training" or "test" set, and the pasts/futures of all the instances align. This can be seen as a special case of multivariate or vector forecasting, with an additional i.i.d. assumption across variables.

The problem covered in this PR is closer to supervised regression in the tabular setting - "past" taking the role of "X" and "future" of "y". 
Individual "pasts" and "futures" across samples may differ, e.g., in the case where relative to a trigger date (e.g., therapy), you have already observed some patients' future (in the training set), while for others you haven't and are looking for a forecast (test/prediction set).

Now it also may be worth to cover panel forecasting, but in my opinion not before we have solved multivariate forecasting properly - since it is more closely related to that.

Quick question: given your view of literature, do you think the above is an accurate reflection? Or am I confusing terms?

partmor:
Hi all, I've been away due to unforeseen issues. I'm getting back on the wagon with your updates and discussions. Will be back to you!

PR #2, opened 2020-11-02
proposal for declarative scityping, checking, conversion of inputs/outputs

fkiraly:
formalized discussion on scityping in a STEP - also compare https://github.com/alan-turing-institute/sktime/pull/420#issuecomment-716059124

Design uses class decorators for implementation - one of the intentions is to make extension to minimize need for implicit knowledge of implementation, and make custom extension of method downwards compatible subject to even substantial changes to "under the hood parts".


mloning (commented):
Yes, thanks for typing this up, I agree with this.

fkiraly:
@mloning, I think this is now a full draft - please review

mloning (approved):
I'm happy with this as a draft, let's merge this! 

I'm not sure about the implementation details. As said, I like the idea of the decorator but we should consider its disadvantages too (e.g. what about debugging and error tracing?).

fkiraly:
>I'm not sure about the implementation details. As said, I like the idea of the decorator but we should consider its disadvantages too (e.g. what about debugging and error tracing?).

It's always about the opportunity cost/benefit - isn't it also quite arcane at the status quo, with two base classes of which one in turn inherits from sklearn `BaseEstimator`?

Further, if the class decorator is robust, there should be no debugging needed for either the extender or the standard user. Think about it like `BaseEstimator` itself - how often did you have to debug `BaseEstimator`?

There might also be an intermediate option if you don't like class decorators:
* make the "general purpose" argument scityping and input/output conversion module
* explicitly use that functionality when extending, or decorate methods instead of the class

fkiraly:
PS, @mloning: if you are in agreement with the user journey but not the implementation details, happy for a counterproposal.

mloning:
@fkiraly I think it's a great idea, but feel like somebody needs to start implementing it to fully understand what it would involve.

PR #4, opened 2020-12-03
multivariate data container discussion

davidbp:
As discussed in  #432 I would like an extension of the package to deal with multivariate data.
This pull requests adds a description of two possible containers for multivariate data.

I like more the idea to store the data as pd.DataFrame object in the case of multivariate timeseries and it could be a sensible way to set a distinction in the fit/predict methods 

- isinstance(input, pd.Series)         -> Univariate case
- isinstance(input, pd.DataFrame) -> Multivariate case

If anyone could give a hand I would like to adapt for example the NaiveForecaster to deal with multivariate data. From there adding other methods should be straightforward. Nevertheless, I don't want to spend much time adapting a model that ends up using an API that does not reach  a good consensus between users/developers of sktime.

### DIFF
--- a/steps/08_supervised_multivariate_forecasting_api/step.md
+++ b/steps/08_supervised_multivariate_forecasting_api/step.md
@@ -1,0 +1,137 @@
+# Multivariate forecasting
+
+Contributors: @davidbp
+
+This proposal describes a design for an updated supervised forecasting module, or an additional module `multivariate`.  It specifies a template class signature for multivariate forecasters.
+
+### Multivariate time series
+
+A Multivariate time series can be represented as a matrix from $\mathbb{R}^{P\times T}$ where $T$ is the "length" of the timeseries and $P$ is the number of features at each time step. In this type of time series we asume that  each variable depends not only on its past values but also has some dependency on other variables. This dependency is used for forecasting future values. 
+
+### Multivariate (or vector) Forecasting
+
+Multivariate Forecasting describes the learning task in which we want to make temporal forward predictions of a given multivariate time series.
+
+### Example of vector forecasting: Vector Auto Regression 
+
+Vector auto regression (or VAR) is one of the most straight forward techniques to do multivariate time series forecasting.
+
+In VAR each variable is a linear combination of the past values of all the other variables. 
+
+In order words,  a vector of values   $\textbf{y}^t = (y_1^t,\dots,y_p^t) \in  Y_1^t \times \dots \times  Y_p^t$  is  predicted from the previous values of the same variables  as follows:  $\textbf{y}^t \approx \textbf{W} \textbf{y}^{t-1} +\textbf{b} $ where $ \textbf{W}$ is a matrix of weights and $\textbf{b}$ is a vector. 
+
+This type of models can be generalized as VAR(K) where the model takes into account the past $K$ time steps of  the timeseries
+$$
+\textbf{y}^t \approx \textbf{W}^1 \textbf{y}^{t-1} +\textbf{W}^2 \textbf{y}^{t-2} + \dots, +\textbf{W}^K \textbf{y}^{t-K}+ \textbf{b}
+$$
+
+
+## API design
+
+Sktime currently has the capabilities for forecasting in univariate timeseries. The multivariate (or vector) forecasting design should be concordant with the previous API for univariate time series. In particular:
+
+- "apply forecaster by row" should be close to the forecasting interface
+- "forecast one time point" should be  close to the supervised time series regression interface
+
+There are some issues with multivariate data, the univariate API uses pd.Series objects, but those are meant to represent one dimensional ndarray objects (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.html). Therefore there are 2 options
+
+- a) Use pandas dataframes where columns indicate features
+
+- b) Use pandas series created with tuples
+
+
+
+#### a) pandas dataframes where columns indicate features
+
+A sensible way to store multiple features is to use a dataframe, as the following example suggests
+
+```
+y = load_airline()
+y_train, y_test = temporal_train_test_split(y, test_size=36)
+multivariate_values  = np.vstack( (y_train.values,np.ones(len(y_train.values)))).T
+df_multivariate_values       = pd.DataFrame(multivariate_values, columns=["f1","f2"])
+df_multivariate_values.index = y_train.index
+print(df_multivariate_values0
+```
+
+```
+            f1   f2
+Period             
+1949-01  112.0  1.0
+1949-02  118.0  1.0
+1949-03  132.0  1.0
+1949-04  129.0  1.0
+1949-05  121.0  1.0
+...        ...  ...
+1957-08  467.0  1.0
+1957-09  404.0  1.0
+1957-10  347.0  1.0
+1957-11  305.0  1.0
+1957-12  336.0  1.0
+```
+
+This type of data container though is not accepted by sktime
+
+```
+forecaster = vector.NaiveVectorForecaster(strategy="last", sp=12)
+forecaster.fit(df_multivariate_values)
+```
+
+```
+AttributeError: 'NaiveVectorForecaster' object has no attribute '_set_y_X'
+```
+
+#### b) pd.Series with tuples
+
+The following example shows one possible solution to represent multivariate data
+
+```
+y = load_airline()
+y_train, y_test = temporal_train_test_split(y, test_size=36)
+multivariate_values  = np.vstack( (y_train.values,np.ones(len(y_train.values)))).T
+y_train_multivariate = pd.Series(list(multivariate_values))
+y_train_multivariate.index = y_train.index
+y_train_multivariate
+```
+
+```
+Period
+1949-01    [112.0, 1.0]
+1949-02    [118.0, 1.0]
+1949-03    [132.0, 1.0]
+1949-04    [129.0, 1.0]
+1949-05    [121.0, 1.0]
+               ...     
+1957-08    [467.0, 1.0]
+1957-09    [404.0, 1.0]
+1957-10    [347.0, 1.0]
+1957-11    [305.0, 1.0]
+1957-12    [336.0, 1.0]
+Freq: M, Length: 108, dtype: object
+```
+
+Note that the code in sktime is not prepared for this a simple object NaiveForecaster fails at predict time. A simple example:
+
+```
+forecaster = NaiveForecaster(strategy="last", sp=12)
+forecaster.fit(y_train_multivariate)      # this does not break current code
+forecaster.predict(fh)                    # this fails
+```
+
+Outputs:
+
+```
+TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
+```
+
+
+
+## Design principles
+
+We wish to adhere to multiple principles:
+
+- consistency with existing template interfaces in sktime, maybe it would be resonable to respresent multivariate data with a dataframe and methods that are passed a dataframe should understand that they deal with multivariate data.
+- avoiding user frustration - natural user expectations on interface behaviour should be met (here separating multivariate and univariate data should be very easy to understand)
+
+
+


mloning:
Thanks @davidbp - I'll take a look over the next few days!

mloning (commented):
Hi @davidbp, 

Thanks again for working on this! 

The `pd.DataFrame` solution with rows representing time points and columns representing variables is my preferred solution for the data container for the multivariate target variable. This is how we currently represent multivariate series (e.g. in the `SeriesToSeriesTransformer`). Adding another representation will further complicate interoperability of our existing functionality and increase complexity for users.

For the API design, I can see two options: 
* we can re-use the univariate forecasters currently implemented in `sktime/forecasting` and dispatch on the input type if necessary (`pd.Series` for univariate and `pd.DataFrame` for multivariate),
* we can create a new framework (e.g. `sktime/multivariate_forecasting`).

Intuitively, I prefer the second option. The interface should be largely the same though. To re-use univariate forecasters, we can provide a reduction estimator that applies separate univariate forecasters to each series of the multivariate target series as a simple solution strategy to the multivariate forecasting problem. We may also be able to refactor other code, however having two types (`Series` and `DataFrames`) may complicate a few things.  

In the STEP, we should also mention that there are three closely related learning tasks to provide a bit more context:
* multivariate (or vector) forecasting
* panel forecasting 
* supervised forecasting #1 

To avoid ambiguity, we should also try to agree on a name for this problem, "vector forecasting" is a bit shorter and more widely used in the econometrics/stats community as far as I can see, "multivariate forecasting" is more prominent in the ML community. I'd go for "multivariate forecasting" - but don't have a strong preference.

I'm happy to push my suggestions directly to your draft if you don't mind.

@fkiraly also wanted to take a look some time soon I believe.

davidbp:
Thanks for your feedback,

Feel free to push any suggestions directly to the draft.

I would like some help adapting a simple model for vector forecasting. Does the current API block the `fit` and `predict` method if the input is a dataframe?

mloning:
Okay - I'll try to add them over the next few days. I'm happy to jump on a call some time this coming week if you're free. @fkiraly may be interested in joining too?

davidbp:
No problem, we can try to talk  almost anytime during this week.

mloning:
@davidbp shall we merge this as work-in-progress and then add to it when we make progress?

davidbp:
Sorry for my late reply, I've done a prototype that does what we commented with respect to forecasting independently per ferature, but the prototype breaks the API that sktime/sklearn use.
For example, imagine you want to perform crossvalidation in a Reducer.

We might have the same model copied `n_features` times (one for each  univariate timeseries).
The API wants the same hiperparameter to be tested for each of the models inside the reducer, but the sensible way would be to create a reducer and allow independent hyperparameters per feature.

How could we do this? Maybe we could have another chat to discuss some of this issues.

davidbp:
I think the previous example did not work because the grid code in sktime checks if the data is a `pd.Series`:

https://github.com/alan-turing-institute/sktime/blob/3f132219e5217ba3dba4b0237f15ea847a84d5f5/sktime/forecasting/model_selection/_tune.py#L169

For example, the `_score` function is:

```
def _score(y_test, y_pred, scorer):
    """Evaluate forecasts"""
    if not isinstance(y_pred, pd.Series):
        raise NotImplementedError(
            "multi-step forecasting horizons with multiple cutoffs/windows "
            "are not supported yet"
        )

    # select only test points for which we have made predictions
    if not np.all(np.isin(y_pred.index, y_test.index)):
        raise IndexError("Predicted time points are not in test set")
    y_test = y_test.loc[y_pred.index]

    scores = {name: func(y_test, y_pred) for name, func in scorer.items()}
    return _check_scores(scores, scorer)
```

mloning:
Hi @davidbp, thanks for the update. 

So there are two questions:
* How should we generalise GridSearchCV and other composition classes? Should they work for univariate and multivariate forecasting? Or should there be separate classes for the separate tasks?
* How can we refer to hyper-parameters of component algorithms in a composite if the components haven't been specified at construction? We could add a convention, e.g. numbering: `forecaster1__<param>`. Or we could allow users to specify a forecaster for each variable similar to how the `ColumnTransformer` works. I'm actually not sure how the `ColumnTransformer` handles the hyper-parameters of the different transformers.

@fkiraly any patterns/ideas that come to mind?

fkiraly:
Hm, very interesting discussion! Thanks for looking into this, @dabidbp.

*Regarding `GridSearchCV`:*

I think the behaviour for multivariate should be "one parameter across all variables", and you optimize by some multivariate prediction loss (e.g., overall MSE, or normalized MSE averaged across all variables). If you want to specify whether you want to tune per variable or overall, I'd split this up by defining an "apply-per-variable" component: it takes any forecaster, and applies it independently to each variable - not sure about the name.

Then you would have a difference between;
* `ForecastCV(my_estimator)`, which tunes over all the variables, and 
* `ApplyPerVariable(ForecastCV(my_estimator))`, which tunes per variable.

Alternatives which I find less nice, since less explicit in specification:
* forecast-CV has two modes: per-variable, and overall
* there are two classes, one for each mode


*Regarding parameters of composites:*

@mloning, why are these hyper-parameters? It sounds more like these are *fitted parameters*, so it would be an question about how we do inspection of those, in a case where there are multiple copies of an internal estimators, like in an abstract bagging wrapper, and you wanting to inspect individual trees in a random forest, no?

Irrespectively, I think your question is moot: I think all hyper-parameters should already be exposed at construction, and that should in return be a definition of what is a "hyper-parameter". Everything else is a fitted parameter, internal variable, or state parameter.

mloning:
So for a heterogeneous composition with a different univariate forecaster for each series, we could write something like this: 
```python
mf = ReducedMultivariateForecaster([
    ("uf1", ARIMA(), [0, 1]), # (<name>, <forecaster>, <column>)
    ("uf2", ExponentialSmoothing(), [3])
    ("uf3", Theta(), [4])
])
``` 
where the column refers to the univariate component series.

fkiraly:
>So for a heterogeneous composition with a different univariate forecaster for each series, we could write something like this:

Yes! Though I would just call it `MultivariateForecaster`, and you should also be specify columns by name(s).

The "names" spec is important, but that would make it the same interface as `sklearn`'s `Pipeline`.

davidbp:
> ```python
> ReducedMultivariateForecaster
> ```

How could we allow this `ReducedMultivariateForecaster`  to train hyper-parameters separately?
We still have the issue that the API checks that the input data is a `pd.Series` object.

Should we then check if the type is a `pd.Series` or a `pd.DataFrame` and make specific code for each option?

Note that there are many functions such as 
```
sktime.utils.validation.forecasting.check_y_X
```
that seem to be created having in mind that the data would be allways a `pd.Series`.

I really like this format:
```
mf = ReducedMultivariateForecaster([
    ("uf1", ARIMA(), [0, 1]), # (<name>, <forecaster>, <column>)
    ("uf2", ExponentialSmoothing(), [3]),
    ("uf3", Theta(), [4])
])
```
I can try to make a prototype and we iterate over it.

fkiraly:
@davidbp, I re-read the entire discussion and I feel that especally the "multiple input/output type" problem and your initial "isinstance" checks relate quite a bit to this suggestion of mine on handling multiple inputs or outputs:
https://github.com/sktime/enhancement-proposals/blob/master/steps/05_scitype_based_IO_checks/step.md

Comments would be appreciated on how it relates to this, or on STEP05 itself, @davidbp.

mloning:
@davidbp You can train hyper-parameters separately using scikit-learn's nested hyper-parameter interface and double-underscore naming convention. 
```python
mf = ReducedMultivariateForecaster([
    ("uf1", ARIMA(), [0, 1]), # (<name>, <forecaster>, <column>)
    ("uf2", ExponentialSmoothing(), [3]),
    ("uf3", Theta(), [4])
])
```
For example, you can specify a grid for parameters for the `ARIMA` model as 
```python
param_grid = {"uf1__d": [0, 1, 2]}
``` 
I think that composition functionality like `GridSearchCV` shouldn't really check input data anyway and instead defer those checks to its components. So I'm happy to make changes to `GridSearchCV`.

mloning:
@fkiraly I don't want to just call it `MultivariateForecaster` because it's a reduction approach. But the reduction approaches are quite wordy for class names. I think we need to think about some more convenient naming scheme or some more user-friendly factory functions to instantiate reduction estimators. 

I'm thinking of something like this but not sure exactly: 
```python
Estimator = get_reduction(from="multivariate-forecasting", to="univariate-forecasting", ...)
estimator = Estimator(...)  # instantiate and specify hyper-parameters 
```

fkiraly:
>But the reduction approaches are quite wordy for class names. I think we need to think about some more convenient naming scheme or some more user-friendly factory functions to instantiate reduction estimators.

I know!

In `sktime` (and some non-public projects) we keep ending up with names like `ReductionForecastingToClassificationByParametricSlidingWindow` or similar.

This makes user specification hard, and shorthands (e.g., scitype based) only help a bit.

Regarding factory, I'm not sure whether this is the most user friendly solution - @lovkush-a and I coincendentally already had a discussion today about sense and usefulness (e.g., for default user) of factory methods like `ReducedTimeSeriesRegressionForecaster`, and I encouraged him to open a discussion in the forum.

davidbp:
Following the convention
```
mf = ReducedMultivariateForecaster([
    ("uf1", ARIMA(), [0, 1]), # (<name>, <forecaster>, <column>)
    ("uf2", ExponentialSmoothing(), [3]),
    ("uf3", Theta(), [4])
])
```
I tried the following class

```
class ReducedMultivariateForecaster:
    """
    Class for using several copies of the same univariate forecaster 
    for each of the columns in the input data.
    
    This class reduces the multivariate forecasting problem to an
    `n_variables` forecasting problem.
    """
    
    def __init__(self, forecasters):
        self.forecasters = forecasters
        
    def fit(self, Y, X=None, fh=None):
        
        assert isinstance(Y,pd.DataFrame),\
            f".fit recieved type(Y)={type(Y)} but was expecting a pd.DataFrame"
            
        self._check_variable_coverage(Y)
        
        self.n_variables = Y.shape[1]
        self.forecasters_ = []
        self.last_index = Y.index[-1]
        self.columns = Y.columns
        
        for i, col in enumerate(Y):
            forecaster = self._get_forecaster(i)
            f = copy.deepcopy(forecaster)
            f.fit(Y[col])
            self.forecasters_.append(f)
                 
        self._is_fitted = True
        return self
            
    def predict(self, fh=None, X=None):
        y = np.empty((len(fh), self.n_variables))
        
        for i, f in enumerate(self.forecasters_):
            y[:, i] = f.predict(fh)
                    
        index_ = pd.RangeIndex(start=self.last_index,stop=self.last_index+len(fh),step=1)
        df_pred =  pd.DataFrame(y, index= index_,columns=self.columns) 
       
        return df_pred
    
    def update(self, Y, X=None):
        # online learning
        for f in enumerate(self.forecasters_):
            f.update(y, X=None)
            
        return self

    def get_params(self, deep=True):
        """
        get_params(deep=True) method of sktime.forecasting.Reducer instance
        Get parameters for this estimator. That is, a list of parameters for each estimator 
        in the .forecasters_ list.

        Parameters
        ----------
        deep : bool, default=True
            If True, will return the parameters for this estimator and
            contained subobjects that are estimators.

        Returns
        -------
        params : list of mapping of string to any
                 List containing dicts with parameter names mapped to their values.

        """
        #params = []
        #for forecaster in self.forecasters_:
        #    params.append(forecaster.get_params(deep=True))
        
        params = {}
        for i, forecaster in enumerate(self.forecasters_):
            params[str(i)] = forecaster.get_params(deep=True)
        
        return params
    
    def _check_variable_coverage(self, Y):
        """
        Checks that self.forecasters cover all the features in Y.        
        """
        n_variables = Y.shape[1]
        
        variable_ids = set()
        for tuple_ in self.forecasters:
            variable_ids = variable_ids.union(tuple_[-1])
            
        assert set(range(n_variables))==variable_ids, "the forecasters do not cover the input data variables"
        
            
    def _get_forecaster(self,i):
        """
        Returns the appropiate forecaster for variable `i`
        """
        for tuple_ in self.forecasters:
            if i in tuple_[-1]: 
                return tuple_[1]
```       

but I would need the check of pd.Series to be removed if we want to use the `ForecastingGridSearchCV` with this type of class.

I have updated the notebook with an example of this class so you might play with this directly:

https://github.com/davidbp/python_tutorials/blob/master/python_libraries/sktime/03_vector_forecasting_Reducer.ipynb

mloning:
Hi @davidbp, yes do you want to open a PR on the main repo to start the new module for multivariate forecasting?

davidbp:
That's a good idea @mloning, I will try to make a PR soon

fkiraly:
@davidbp, FYI - after a while we've converged here:
https://github.com/alan-turing-institute/sktime/pull/980

This is a bit of a departure, but still in line with the part of your suggestion to use `pd.DataFrame` for multivariate series.
However, we're planning to support multiple input types, so it will be fine to pass one-column `pd.DataFrame` instead of `pd.Series`, for instance.

fkiraly:
Thanks again for your contributions, @davidbp, see more extensive thank you message at the end of https://github.com/alan-turing-institute/sktime/issues/432.

I'm closing this PR as it has been superseded by the related (but different) "doing it in the same base class" design mentioned above.

PR #9, opened 2021-03-19
Adding a STEP for Hierarchical Forecasting.

satya-pattnaik:
# Enhancement Proposal for Hierarchical Forecasting Techniques.
This STEP has a summary of some well known Hierarchical Forecasting Techniques to address Time Series in a Hierarchy.


fkiraly (changes requested):
This is in-principle in scope, but would you mind being more explicit about the software interface and API design?

For example, would it just use the current forecasting interface and is this just adding more models? Or are new interface elements necessary (only then it makes sense to open a step).

Example usage may be a good start.

satya-pattnaik:
> This is in-principle in scope, but would you mind being more explicit about the software interface and API design?
> 
> For example, would it just use the current forecasting interface and is this just adding more models? Or are new interface elements necessary (only then it makes sense to open a step).
> 
> Example usage may be a good start.

I added a brief example usage. Based on the feedbacks I can refine it further.

steps/hierarchical_ts/step.md
```diff
@@ -0,0 +1,117 @@
...
+ts = load_data()
+ts
+```
+![time_series_data](e_prop_1.png)<br>
+
+**Aggregate Data**(For each state i.e CA and TX)
+```
+ts["CA"] = ts["CA_1"] + ts["CA_2"]
+ts["TX"] = ts["TX_1"] + ts["TX_2"]
+ts
```
AngelPone, line 73: I think there should be a class to define the hierarchy structure, because there are many types of hierarchical time series, such as group, sectional, temporal and cross-temporal, the class will be used in hierarchical forecasting models.

steps/hierarchical_ts/step.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+![forecasts](e_prop_3.png)<br>
+
+**Hierarchical Forecast Design Sketch**
+(This will generate the coherent forecasts across the hierarchy)
+
+```
+hierarchical_reconciler = HierarchicalReconcile(method="ols")
+hierarchical_reconciler.fit(ts)
```
AngelPone, line 103: use `forecasts` and the hierarchical structure I mentioned before to fit the model.

steps/hierarchical_ts/step.md
satya-pattnaik: Do you mean something like:

```
class BaseHierarchicalForecaster:

class CrossSectional(BaseHierarchicalForecaster):

class Group(BaseHierarchicalForecaster):

class Temporal(BaseHierarchicalForecaster):
```
Within each class we create hierarchies similar to hts/pyhts,i.e
pass column names as underscore separated strings and use chars = [2,3] to separate?

This is just an initial sketch, we need to refine it further.

satya-pattnaik: If you are fine we can have an initial short call on this.

fkiraly:
>I think there should be a class to define the hierarchy structure, because there are many types of hierarchical time series, such as group, sectional, temporal and cross-temporal, the class will be used in hierarchical forecasting models

@satya-pattnaik, I think @AngelPone is right, but your suggestion is not what he means.

As far as I see, you are trying to encode annotations of variables in a series, or instances in a panel, correct? So you are looking for an abstract description of hierarchies `hier = Hierarchy(type1vars = ["bla", "foo"], type2vars = ["bar"])` that you could use either in a constructor, such as 

`myhierarchicalforecaster = HF(params, hierarchy_desc=hier)`,

or in fit/predict, such as

`myhierarchicalforecaster.fit(y, X, hierarchy_desc)`?

If so, I tend towards passing in the constructor, because this considers the hierarchy information as a part of the model rather than intrinsic to the data.

But indeed, I think a call and a good write-up of use cases is needed to get us further here.

AngelPone:
Here is a design draft and use case as I suggested:



```Python
class BaseHierarchy():
    def aggregate_ts():
    # construct df used for generating base forecasts and reconciliation

class CrossSectionalHierarchy(BaseHierarchy):
    # classmethod used for construction of hierarchy from different abstract descriptions.
    @classmethod
    def from_xxx():
        return cls()

class TemporalHierarchy(BaseHierarchy):
    def __init__():
```

Forecasters:

```python
def wls():

def mint():

class HierarchicalForecaster(BaseForecaster):
    def __init__(hierarchy, base_forecasters:Union[BaseForecaster|List|Dict], method="ols", 
                  weights=Optional[str])

    def _check_y_X():

    def _fit_forecasters(y, X, return_residuals=False):
    def _predict_forecasters(X):

    def _fit(y, X):
        df = self.hierarchy.aggregate_ts(y)

        if self.method=='ols':
            self._fit_forecasters(df, X)
            self.G_ = wls(self.hierarchy, weights=None)
        elif self.method == 'wls':
            self.G_ = wls(self.hierarchy, weights=weights)
        elif self.method == 'mint':
            residuals = self._fit_forecasters(df, X, True, weights=weights)
            self.G_ = mint(self.hierarchy, residuals)
        else:
            raise ValueError()

    def _predict(X):
        base_forecasts = self._predict_forecasters(X)
        reconciled_forecasts = self.G_.dot(base_forecasts)
        if isinstance(self.hierarchy, TemporalHierarchy):
            return reconciled_forecasts.flat()
        return reconciled_forecasts
```

AngelPone:
use case:

```Python
y = load_data()
ht = Hierarchy.from_xxx()
hforecasters = HierarchicalForecaster(ht, method="ols", base_forecasters=AutoArima(sp=7))
hforecasters.fit(y)
reconciled_forecasts = hforecasters.predict([1])
```

There should be evaluation and plotting methods designed for `Hierarchy` due to the complexity of hierarchical data(base forecasts, reconciled forecasts and multiple levels, e.g., users may focus on forecasting performance of specific levels or series)

aiwalter:
hi together, I am also quite interested in HTS. Is there an idea how to best continue with the topic? Probably as @fkiraly has proposed, it would be best to have a quick call to finalize this STEP.
what do you think @AngelPone @satya-pattnaik ?

satya-pattnaik:
> hi together, I am also quite interested in HTS. Is there an idea how to best continue with the topic? Probably as @fkiraly has proposed, it would be best to have a quick call to finalize this STEP.
> what do you think @AngelPone @satya-pattnaik ?

Sounds good @aiwalter . A call towards last week of this month will be fine. Can we have a sub channel for this on slack?

mloning:
```python
class Reconciler()
    def fit()
        ...
    def reconcile()
        ...
```

satya-pattnaik:
Pinging @AngelPone, @aiwalter.

Also can someone help me with GitHub username of Leonidas.

fkiraly:
@AngelPone, FYI the decision on Friday was to use your design, basically unchanged.

Leonidas is @ltsaprounis

steps/hierarchical_ts/step.md
```diff
@@ -98,14 +100,98 @@ print(forecasts)
...
+
+class TemporalHierarchy(BaseHierarchy):
+    def __init__():
+
+# Forecasters
+def wls():
+
+def mint():
+
+class HierarchicalForecaster(BaseForecaster):
```
fkiraly, line 134: Do we want an abstract base class for hierarchical forecasters, or at least an extension template?
Not all hierarchical forecasters are reconciliation based.

steps/hierarchical_ts/step.md
```diff
@@ -98,14 +100,98 @@ print(forecasts)
...
+    def __init__(hierarchy, base_forecasters:Union[BaseForecaster|List|Dict], method="ols", 
+                  weights=Optional[str])
+
+    def _check_y_X():
```
fkiraly, line 138: why is this needed?

steps/hierarchical_ts/step.md
```diff
@@ -98,14 +100,98 @@ print(forecasts)
...
+
+    def _fit_forecasters(y, X, return_residuals=False):
+    def _predict_forecasters(X):
+
+    def _fit(y, X):
+        df = self.hierarchy.aggregate_ts(y)
+
+        if self.method=='ols':
+            self._fit_forecasters(df, X)
```
fkiraly, line 147: why is this needed? seems like a one-liner

steps/hierarchical_ts/step.md
```diff
@@ -98,14 +100,98 @@ print(forecasts)
...
+            self.G_ = wls(self.hierarchy, weights=None)
+        elif self.method == 'wls':
+            self.G_ = wls(self.hierarchy, weights=weights)
+        elif self.method == 'mint':
+            residuals = self._fit_forecasters(df, X, True, weights=weights)
```
fkiraly, line 152: why would `_fit_forecasters` compute residuals?

steps/hierarchical_ts/step.md
```diff
@@ -98,14 +100,98 @@ print(forecasts)
...
+                  weights=Optional[str])
+
+    def _check_y_X():
+
+    def _fit_forecasters(y, X, return_residuals=False):
+    def _predict_forecasters(X):
+
+    def _fit(y, X):
+        df = self.hierarchy.aggregate_ts(y)
+
```
fkiraly, line 145: why fo we reimplement instead of interfacing `pyhts`?

steps/hierarchical_ts/step.md
satya-pattnaik: Yes true that,
I will propose something like this - For all of the reconciliation based approaches-
```from sktime.hierarchical.reconcile import HierarchicalReconciler```

And another submodule , for Panel Hierarchical Models-
```from sktime.hierarchical.multilevel import EffectsModel```

steps/hierarchical_ts/step.md
satya-pattnaik: I am **assuming** here that @AngelPone is using ```_fit_forecasters``` for the BaseForecasters fitting, where residuals are computed for base level time series and  then passed on to approaches like ```mint```.

I suggest that the BaseForecaster fitting has to be decoupled from the HierarchicalForecasting module. As a user will like to have some degree of control to fit the Base forecasters and check the goodness of fit(DW,Residual plot etc). In [PyHTS](https://github.com/AngelPone/pyhts/blob/master/pyhts/HFModel.py) as I see it's not the case.

So what I propose, in the context of the following constructor -

```def __init__(hierarchy, base_forecasters:Union[BaseForecaster|List|Dict], method="ols", weights=Optional[str])```

The ```base_forecasters``` are pre-fitted ```Sktime Forecaster objects/pipelines```, which the user fits before moving ahead with Reconciliation.

And instead of ```_fit_forecasters```, we can have ```_predict_residual```. 
Your views @AngelPone?

steps/hierarchical_ts/step.md
satya-pattnaik: Your views on this @aiwalter, @ltsaprounis ??

fkiraly:
superseded by https://github.com/sktime/enhancement-proposals/blob/main/steps/14_global_forecasting/step.md

PR #13, opened 2021-08-02
STEP for transformer refactor

fkiraly:
This is a STEP for the transformer refactor, which will in turn rely on preparations in sktime PR 1125.

FYI @mloning, @TonyBagnall, @aiwalter, @chrisholder, @guzalbulatova, @sveameyer13

### DIFF
--- a/steps/11_transformer_refactor/step.md
+++ b/steps/11_transformer_refactor/step.md
@@ -1,0 +1,158 @@
+# refactor of transformers interface
+
+Contributors: @fkiraly, 
+
+## High-level summary 
+
+### The Aim
+
+The transformers interface has grown organically and is difficult to use, interface, and extend.
+
+This proposal tries to alleviate the problem by removing mixins from transformers and using the new tag system to provide a unified and easy-to-use interface.
+
+### The proposed solution
+
+Our proposed solution follows principles used in the refactor of the forecasters:
+
+* methods are separated into outer `fit`/`transform` and inner `_fit`/`_transform` etc
+* the four mixins are removed, and replaced by tags
+* the four transformer mixins are united under a single interface
+* addition of `update` methods for transformers, to be usable with forecaster `update`
+
+The unification introduces type inhomogeneity in the `transform` method - this is addressed by allowing both `Series` and `Panel` inputs and outputs, and dropping the `Z` argument in favour of `X` (keeping the `Z` as a downwards compatible alias).
+
+## Refactor design: transformers
+
+We proceed outlining the refactor interface.
+
+### Conceptual model: types of transformers
+
+Conceptually, we change our view of transformer types. Before the refactor, these weree:
+
+* series-to-primitives
+* panel-to-tabular
+* series-to-series
+* panel-to-panel
+* series-to-panel
+
+Where series-to-panel occurs only in two places currently and not formalised as a transformer type:
+* sliding window forecasting (in `update_predict`)
+* epoching, in `sktime-neuro`
+
+We replace the old taxonomy with an input/output and rowwise yes/no taxonomy - based on what type an individual instance has, and whether `fit`/`transform` is independent per-instance or not:
+
+| `transform-input` | `transform-output` | `rowwise` | old name |
+|---|---|---|---|
+| `Series` | `Primitives` | True | series-to-primitives |
+| `Series` | `Primitives` | False | panel-to-tabular |
+| `Series` | `Series` | True | series-to-series |
+| `Series` | `Series` | False | panel-to-panel |
+| `Series` | `Panel` | True | N/A, this is new |
+| `Primitives` | `Primitives` | True | unfittable scikit-learn transformer |
+| `Primitives` | `Primitives` | False | fittable scikit-learn transformer|
+
+Implications on the signature will be discussed below, and we will repeat the table after a longer explanation.
+
+### Fit/transform signature
+
+We begin by outlining the expected fit/transform signature for each of the "old" transformer types.
+
+#### series-to-primitives and panel-to-tabular
+
+`fit(X: Panel or Series, y=None) -> self`
+
+`transform(X: Panel or Series, y=None) -> res: pd.DataFrame`
+
+where in `transform`, rows in `res` correspond to instances in `X`.
+If `X` is a `Panel`, `transform` is applied per row; if `X` is a `Series` it is considered as a single-series-panel.
+The exogeneous input `y` can be any label vector but is usually `None` (discussed later).
+
+The distinction between series-to-primitives and panel-to-tabular lies not in the `transform` interface, but in assumptions
+about the statistical relation between input and output:
+* all pairs of input rows of `fit` and output rows of `transform` are assumed statistically independent for series-to-primitives - this is modelled as a boolean property tag `scitype:rowwise = True`.
+* this assumption is not made for panel-to-tabular - this is modelled as a boolean property tag `scitype:rowwise = False`.
+
+Series-to-primitives transformers should be considered vectorized in `fit`, with `Series` being the "native" input and `Panel` being
+a vectorized
+
+#### series-to-series and panel-to-panel
+
+`fit(X: Panel or Series, y=None) -> self`
+
+`transform(X: Panel or Series, y=None) -> res: Panel or Series`
+
+where instances in `res` correspond to instances in `X`; and `X` in `fit` and `transform` must have the same number of instances and variables.
+If `X` is a `Panel`, `transform` is applied per row; if `X` is a `Series` it is considered as a single-series-panel.
+The exogeneous input `y` can be any label vector but is usually `None` (discussed later).
+
+The distinction between series-to-series and panel-to-panel lies not in the `transform` interface, but in assumptions
+about the statistical relation between input and output:
+* all pairs of input rows of `fit` and output rows of `transform` are assumed statistically independent for series-to-series - this is modelled as a boolean property tag `scitype:rowwise = True`.
+* this assumption is not made for panel-to-panel -  this is modelled as a boolean property tag `scitype:rowwise = False`.
+
+#### series-to-panel
+
+`fit(X: Series, y=None) -> self`
+
+`transform(X: Series, y=None) -> res: Panel`
+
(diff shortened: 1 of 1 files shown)


steps/11_transformer_refactor/step.md
```diff
@@ -0,0 +1,158 @@
...
+
+* series-to-primitives
+* panel-to-tabular
+* series-to-series
+* panel-to-panel
+* series-to-panel
+
+Where series-to-panel occurs only in two places currently and not formalised as a transformer type:
+* sliding window forecasting (in `update_predict`)
+* epoching, in `sktime-neuro`
```
SveaMeyer13, line 40: Hi @fkiraly, I'm not entirely sure whether we should handle the epoching as a transformer. (not a strong opinion just bouncing off my thoughts:) What the use case here usually requires is that we can pass a time series and one or more labels, e.g. "right" and "left" and an interval e.g. [-0.5 1]. And then this function/transformer cuts out all the parts of the data where these events occurred. It then returns panel data and a labels vector (describing each instance of the panel). So the return structure would be similar to the one you get when loading a toy dataset in sklearn (it simply returns X and y). And it could then further be used just like a normal classification problem. The critical point I see is that a transformer would not be able to return the labels vector y but would require it to be fixed in the beginning. Which I think is a bit unhandy.

I see two options of forcing this use case into a transformer: 
(1) We could make the user specify only one label. Then the label of each instance in the panel would be clear. And the user would then have to apply this transformer per label, e.g. first cutting out all the "right" trials and then cutting out all the "left" trials. Here we would however have the risk that the user (accidentally) applies different panel-to-panel transformers to the two generated panels afterwards, thereby classifying what happened to the data afterwards rather than purely classifying the label.  
 
(2) We could require the labels vector (y) (indicating which instance of the panel is supposed to be of which label afterwards) to be fixed beforehand. This would basically require a function `create_labels_vector` that creates the labels vector based on the labels we want to cut out and an annotation object (currently a `pd.DataFrame` that contains events and onsets). So instead of calling simply one function we would require the user to first create the labels vector and then to `fit_transform()` the series.

steps/11_transformer_refactor/step.md
```diff
@@ -0,0 +1,158 @@
...
+
+We replace the old taxonomy with an input/output and rowwise yes/no taxonomy - based on what type an individual instance has, and whether `fit`/`transform` is independent per-instance or not:
+
+| `transform-input` | `transform-output` | `rowwise` | old name |
+|---|---|---|---|
+| `Series` | `Primitives` | True | series-to-primitives |
+| `Series` | `Primitives` | False | panel-to-tabular |
```
SveaMeyer13, line 47:
```suggestion
| `Panel` | `Primitives` | False | panel-to-tabular |
```

steps/11_transformer_refactor/step.md
```diff
@@ -0,0 +1,158 @@
...
+| `Series` | `Series` | True | series-to-series |
+| `Series` | `Series` | False | panel-to-panel |
```
SveaMeyer13, line 49:
```suggestion
| `Panel` | `Panel` | False | panel-to-panel |
```

steps/11_transformer_refactor/step.md
```diff
@@ -0,0 +1,158 @@
...
+
+Only the five combinations above are allowed (implemented), plus "primitives-to-primitives" which is the tag combination for scikit-learn type transformers.
+
+To avoid confusion, the old terminology "series-to-series" etc is no longer used, instead we talk about input/output types.
+The following table gives a translation between old and new scheme.
+
+| `transform-input` | `transform-output` | `rowwise` | old name |
+|---|---|---|---|
+| `Series` | `Primitives` | True | series-to-primitives |
+| `Series` | `Primitives` | False | panel-to-tabular |
```
SveaMeyer13, line 117:
```suggestion
| `Panel` | `Primitives` | False | panel-to-tabular |
```

steps/11_transformer_refactor/step.md
```diff
@@ -0,0 +1,158 @@
...
+| `Series` | `Series` | True | series-to-series |
+| `Series` | `Series` | False | panel-to-panel |
```
SveaMeyer13, line 119:
```suggestion
| `Panel` | `Panel` | False | panel-to-panel |
```

steps/11_transformer_refactor/step.md
fkiraly: This is how I see it would be part of the framework:

* it would be a "series-to-panel" transformer which takes a `y` that is of segmentation scitype.. 
* the segments (in your diction: intervals, e.g., [-0.5, 1]) would be contained in `y`, e.g., as an `IntervalIndex` or in some other encoding. 
* the series would be passed in `X`. The return of transform is a panel which contains all the "cut out" sub-sequences.

fkiraly: so sth like this

```python
mytrafo = SegmentCutter()
y = pd.IntervalIndex.from_arrays([-0.5, 10, 14], [1, 12, 15])
epochs = mytrafo.fit_transform(X, y)
```

now `epochs` is a panel which contains cut parts form `X`, in intervals [-0.5, 1], [10,12], and [14,15].

steps/11_transformer_refactor/step.md
fkiraly: no, no - this is right and intended like this!
The `Series` refers to "what is in one row". 
Multiple instances of `Series` make a `Panel`, and multiple instances of `Primitives` make a table (doesn't have a type).

steps/11_transformer_refactor/step.md
fkiraly: also this is intended the way it is

steps/11_transformer_refactor/step.md
fkiraly: of course, if you think it *should* be different from what I intended, please let me know (I think there are arguments for that too).
The reply is based on me thinking that you were thinking this was an unintended mistake rather than explicit intention on my side.

fkiraly: come to think of it, maybe the explanation of what I was thinking could be improved, too?

steps/11_transformer_refactor/step.md
SveaMeyer13: But how would you know the labels of the instances in epochs? You'd have to have an additional function that creates the labels vector from y in order to train a classifier, right?

steps/11_transformer_refactor/step.md
```diff
@@ -0,0 +1,158 @@
...
+
+We replace the old taxonomy with an input/output and rowwise yes/no taxonomy - based on what type an individual instance has, and whether `fit`/`transform` is independent per-instance or not:
+
+| `transform-input` | `transform-output` | `rowwise` | old name |
+|---|---|---|---|
+| `Series` | `Primitives` | True | series-to-primitives |
+| `Series` | `Primitives` | False | panel-to-tabular |
+| `Series` | `Series` | True | series-to-series |
+| `Series` | `Series` | False | panel-to-panel |
+| `Series` | `Panel` | True | N/A, this is new |
```
SveaMeyer13, line 50: I now understand how the above was intended. But then why do you have `panel` here?

fkiraly: well, because this is the only case where an individual series gets "changed" into a panel, i.e., one series input corresponds to one panel output.

In all the other cases, you have one-to-one correspondence between individual series and/or dataframe rows, on the index level.

steps/11_transformer_refactor/step.md
fkiraly: ah, I see - you have labels!
Well, in that case, you would start with a labelled segmentation, let's say `Y`, which is a `pd.Series` with a `pd.IntervalIndex`, and entries are labels.
You could now pass `y=Y.index` to the `SegmentCutter` as above, and `Y.reset_index()` would give you the labels - unless you allow panels which can also have a `pd.intervalIndex` as instance index, in which case `epochs` and `Y` already would have the same index and could be used in a classifier.

mloning (commented):
Hi @fkiraly, I had a quick look.

I'm in favour of the refactoring along the lines of the forecasting module with public and inner fit methods. 

The proposed change in `fit(X: Panel or Series, y=None) -> self` is a big interface change. I think we need to move carefully here and see how that works on in terms of efficiency and maintainability. I'm not aware of a way of writing a general rowwise applications of a transformation without looping in Python which would be very inefficient.

fkiraly:
>I'm not aware of a way of writing a general rowwise applications of a transformation without looping in Python which would be very inefficient.

`numpy.apply_over_axes`, `numpy.apply_along_axis`, `pandas.DataFrame.apply`, and - in case of many transformations - making used of vectorized operations in `numpy` or `pandas`.

Many algorithms can be implemented, and in many cases of the interfaced algorithms, are already implemented in vectorized fashion.

>The proposed change in fit(X: Panel or Series, y=None) -> self is a big interface change

It is, but we can make it so it is fully downwards compatible - it would just add to the series transformers the capability to run on panels. Currently, you have to iterate over rows, so the current way of doing this is very inefficient.

mloning:
> numpy.apply_over_axes, numpy.apply_along_axis, pandas.DataFrame.apply, and - in case of many transformations - making use of vectorized operations in numpy or pandas.

As far as I'm aware, in many cases, they're not much more efficient than for loops and you don't get fitted parameters back from them when applying `fit_transform`.

fkiraly:
>you don't get fitted parameters back from them when applying `fit_transform`.

What do you mean by that? Why would you even expect to get fitted parameters when applying `numpy.apply_etc`? Or, what are you comparing to?

mloning:
What I'm thinking about is this: When you apply a transformer to each row via `fit_transform`, you may not only want to transform the data but also keep the fitted parameters for each row-wise application. I'm not sure how this can be achieved in an efficient way. 

With the refactoring, is it still possible to write different implementations for series-to-series and panel-to-panel transformers? 

I haven't had much more time to look at the proposal in more depth. My general feeling is that we're moving a bit too quickly at the moment, we're introducing many API changes without giving them much time to trickle through to all modules. Not sure if I'm the only one who thinks that?

PR #15, opened 2021-09-10
STEP: Probabilistic forecasting

fkiraly:
I made a STEP for a full probabilistic forecasting interface, see also https://github.com/alan-turing-institute/sktime/issues/984.

FYI, @mloning, @aiwalter, @ltsaprounis, @sveameyer13

### DIFF
--- a/steps/12_proba_forecasting/step.md
+++ b/steps/12_proba_forecasting/step.md
@@ -77,7 +77,9 @@ Each new method has a private counterpart which is called internally, i.e., `_pr
 
 To ensure downwards compatibility, the `return_pred_int` and `alpha` parameters will still be accepted by `predict` until the end of the next deprecation cycle.
 
-If these are passed and `return_pred_int=True`, the base class directs the `predict` call to `predict_interval` where possible; for downwards compatibility, a temporary boolean tag `old_predict_interval_logic` directs to `_predict` instead, with the aim to remove prediction interval 
+If these are passed and `return_pred_int=True`, the base class directs the `predict` call to `predict_interval` where possible; Note: this is always possible once a specific forecaster is refactored. Whether a forecaster is already refactored will be checked for in an additional class method `_has_refactored_predict_interval()`;
+
+Since the return type of prediction intervals also changes, a boolean tag `keep_old_return_type=True` will temporarily be passed to `predict`, allowing the user to decide whether he wants to use the old return type (also possible until the next deprecation cycle". 
 
 ### removal/refactor of unnecessary `BaseForecaster` methods
 


steps/12_proba_forecasting/step.md
```diff
@@ -0,0 +1,87 @@
...
+
+* `predict_var(self, fh=None, X=None, cov=False) -> np.ndarray` returns: if `cov=False`, an vector of same length as `fh` with predictive marginal variances; if `cov=True`, a square matrix of size `len(fh)` with predictive covariance matrix
+
+* `predict_interval(self, fh=None, X=None, alpha=Union[float, np.ndarray[float, dim=1]]) -> pd.DataFrame` returns a `pd.DataFrame` with rows having `fh` as index, and columns having name `[varname]_P[z]` where `varname` are variable names in `y` (passed to `fit`), and `z` are requested quantiles (in percentage, as float-print, without the `%` symbol). If `y` has no variable names (e.g., a `Series` without `name` attribute), the column names are `P[z]` instead. Entries are the predictive quantile for the horizon in the row, for the variable with `varname` at the quantile `z`, as in the column. If `alpha` is a float, then the returned quantiles are `alpha`, `0.5` and `1-alpha` for all variables in `y`. If `alpha` is a vector, then the returned quantiles are exactly the elements of `alpha`, for all variables in `y`.
```
SveaMeyer13, line 47: What should the return type look like for a multivariate series? Would we just add more columns or would it make sense to have a list of pd.DataFrames (one pd.DataFrame for each univariate series/variable) or some kind of nested pd.DataFrame?

fkiraly: well, what I thought it should be is spelled out - just multiple columns, with the naming convention `[varname]__P[z]`. Assuming you want 90% confidence bands and pass `alpha=0.1`, and your original `y` had column names `foo`, `bar`, your return will have six columns, named `foo__P10`, `foo__P50`, `foo__P90`, `bar__P10`, `bar__P50`, `bar__P95`

fkiraly: a nicer alternative could be having a column multi-index with first level being variable name, second level being a float index with values from `alpha`? That would allow to display and index non-integer percentages easily.

Perhaps that's the better way?

SveaMeyer13: I like the second option better. Seems more structured to me and it probably makes it easier to plot/process it afterwards.

fkiraly: ok - I've changed the STEP to reflect this and added an example.

steps/12_proba_forecasting/step.md
```diff
@@ -0,0 +1,89 @@
...
+
+### public/private interface
+
+Each new method has a private counterpart which is called internally, i.e., `_predict_var`, `_predict_interval`, `_predict_proba`. As in the generic `sktime` design, the public method contains "plumbing", while the private class is the extension locus and contains only method logic.
+
+### Downwards compatibility: `predict`
+
+To ensure downwards compatibility, the `return_pred_int` and `alpha` parameters will still be accepted by `predict` until the end of the next deprecation cycle.
+
+If these are passed and `return_pred_int=True`, the base class directs the `predict` call to `predict_interval` where possible; for downwards compatibility, a temporary boolean tag `old_predict_interval_logic` directs to `_predict` instead, with the aim to remove prediction interval
```
SveaMeyer13, line 80:
```suggestion
To ensure downwards compatibility, the `return_pred_int` and `alpha` parameters will still be accepted by `predict` until the end of the next deprecation cycle.

If these are passed and `return_pred_int=True`, the base class directs the `predict` call to `predict_interval` where possible; Note: this is always possible once a specific forecaster is refactored. Whether a forecaster is already refactored will be checked for in an additional class method `_has_refactored_predict_interval()`;

Since the return type of prediction intervals also changes, a boolean tag `keep_old_return_type=True` will temporarily be passed to `predict`, allowing the user to decide whether he wants to use the old return type (also possible until the next deprecation cycle". 
```

SveaMeyer13 (commented):
Hi @fkiraly, I updated the deprecation logic.  

Once we find an agreement on the exact return type of predict_interval, I'll also update that section accordingly. 

Should testing also be discussed in this STEP?

steps/12_proba_forecasting/step.md
fkiraly: makes sense, thanks!

fkiraly:
>Should testing also be discussed in this STEP?

Yes, it should be a section. Do you want to make a stab at it?

PR #16, opened 2021-09-27
Design Time Series Classification/Regression Module

RavenRudi:
Hi @fkiraly,

here is the enhancement proposal as requested. There is still plenty of room for discussion, so please let me know what you think.


fkiraly:
I think this has now been superseded by the current regression module and the estimator design.

We still need to think about common parent classes, but let's revisit that later.

PR #17, opened 2021-10-15
STEP 14: hierarchical, global, and panel forecasting

fkiraly:
This PR introduces a STEP for global forecasting.

### DIFF
--- a/steps/14_global_forecasting/step.md
+++ b/steps/14_global_forecasting/step.md
@@ -30,6 +30,7 @@ Below, variables `y` and `X` can be hierarchical time series.
 For the prototype, there will be only one `mtype` representing hierarchical time series, as follows.
 
 A hierarchical time series is represented by a `pd.DataFrame`, which has a multi-index of `n` levels, `n` being 1 or larger. The last, `n`-th level of the index must be one of the supported temporal indices in `sktime` (see data type specification). The other levels can have arbitrary values.
+Below, we call these `n-1` levels "non-temporal".
 
 Post-prototype, further `mtype`-s may bei ntroduced.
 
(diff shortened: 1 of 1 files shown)


danbartl:
@fkiraly I thought @aiwalter mentioned that someone might contribute their hierarchical reconciliation library to this end? 

In any case, for step 2 we would need the adaptation of one of those methods:  https://github.com/carlomazzaferro/scikit-hts or https://github.com/mavillan/mahts to allow us to pass:

a set of historical values
a set of forecast values
a structure defining the hierarchy
an argument specifiying the hierarchical reconcilation method (top down optimal etc.)

to a method that would do the hierarchical reconciliation.

ltsaprounis:
> 1. **Forecasting using hierarchical information as features** can be simply done by any forecaster that can use exogeneous information
> 2. **hierarchical reconcilation** is not invariably linked to global forecasting. Some use cases will need it, some dont. And hierarchical reconcilation can also be applied to many problems other than global forecasting

I agree with the above, hierarchical features are just features. Based in the image below they can be considered as static covariates which is perfectly captured by the introduction of Z! 
forecasting_framework
image from: [Lim, B.; Arık, S.Ö.; Loeff, N.; Pfister, T. Temporal Fusion Transformers for Interpretable Multi-Horizon Time Series Forecasting. Int. J. Forecast. 2021](https://arxiv.org/pdf/1912.09363.pdf)

However, I like the idea of allowing for multiple levels in the multiindex where their combination defines a single time series instead of a unique time series id. This future-proofs the data container for the hierarchical/grouped case, but it's not a critical feature for global forecasting.

PR #20, opened 2022-05-19
STEP for pyWATTS enhancement

benHeid:
#### What does this implement/fix? Explain your changes.
This STEP proposes to cooperate with pyWATTS to enable a graph pipeline in sktime.

### DIFF
--- a/steps/16_pywatts_graph_pipeline/step.md
+++ b/steps/16_pywatts_graph_pipeline/step.md
@@ -44,15 +44,46 @@ such pipelines in pyWATTS.
 * How is the data flow be realised
   * Step fetches the data and maintains a buffer
 
-## Description of proposed solution
+## Design considerations
 
-### Adding transformers, etc. to a graph pipeline
-#### Requirements:
-1. It has to be possible to add a transformer with multiple inputs
-2. It has to be possible to use a transformer as input for multiple other transformers
-3. Passing additional parameters to control the execution of the transformers etc.
+### Requirements
 
-#### Solution 1: Current Keras-Style API Solution in pyWATTS
+1. it should be possible to use the pipeline as an `sktime` estimator
+2. It has to be possible to add a estimators with multiple inputs
+3. It has to be possible to use a transformer as input for multiple other estimators
+4. parameters to control the execution of the transformers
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
I think this is a great start!
This is definitely more than sufficient for discussion next weeks.
Kindly let me know where you would like to have input.

Some points that I see here:

* most importantly, there is a bit of a discrepancy between more pywatts-sided usage designs and the `sktime` architecture. For instance, `sktime` estimators have different core API methods, such as `fit`/`transform` (transformer) and `fit`/`predict` (forecasters, classifiers), not `train`. How the two map onto each other might be worth discussing.
* maybe it would be a good idea to paste the open design questions we identified from the mardown somewhere in here, e.g., at the end?
* requirements refer only to transformers. Are we restricting ourselves only to transformers at the moment? I think where it becomes "really interesting" is once estimators of specific type, e.g., forecasters, are involved.

fkiraly:
Question FK - what is the difference between solution 1 and 3?

fkiraly:
question: in solution 3, is the `Transformer` one single class that is always used, or would it be a specific (concrete) sktime transformer?

fkiraly:
question: in solution 3/4, you write `train`, I assume if we use the `sktime` transformer interface, this would be `fit` or `fit_transform`?

fkiraly:
looks like current pywatts design, square bracket dunder is like a `ColumnSelectionTransformer`

benHeid:
> * designs and the `sktime` architecture. For instance, `sktime` estimators have different core API methods, such as `fit`/`transform` (transformer) and `fit`/`predict` (forecasters, classifiers), not `train`. How the two map onto each other might be worth discussing.

Solution 1 is the current pyWATTS solution. I.e., the pipeline has to be initialised before the structure of the pipeline is specified. Here, the column selection is done by the square brackets dunder. Solution three is a full keras API. Where the pipeline is initialisied with the inputs and the outputs after the structure is defined. This solution would use an other mechanism for selecting the input columns. Here I proposed something like `Column("bar")` which is in some way analogue to the InputLayer of Keras

fkiraly:
solution 4 looks like it has a typo - `step_information` should all be `graph_pipeline`

PR #21, opened 2022-06-03
Frequency horizon and frequency

khrapovs:
Related issue: https://github.com/alan-turing-institute/sktime/issues/1750


fkiraly (approved):
I think these are good options for discussion

PR #22, opened 2022-06-05
STEP 17: proposed additions to `ForecastingHorizon` handling of `freq`

fkiraly:
Proposed additions to the `freq` STEP.

Adds more explanation of current state and post-deprecation state.

Adds another option of type 2, and an idea how to handle multiple inputs to extract `freq`.


khrapovs (approved):

khrapovs:
I would merge, but I do not have write access...

fkiraly:
now you do, sorry for not giving that to you.
Also see https://github.com/alan-turing-institute/sktime/issues/2753, you not having write access has highlighted the problem to me.

PR #23, opened 2022-06-22
Add step: CV with irregular TS data

khrapovs:
Add step: CV with irregular TS data


fkiraly (approved):
Thanks, makes sense!

fkiraly:
(approval is not an approval of the idea, just that the STEP is "sufficiently complete")

PR #24, opened 2022-07-02
[ENH] index formatted cutoff option for STEP 17 (dealing with `freq` deprec)

fkiraly:
This adds an option 3 to STEP 17 on how to deal with the `freq` deprecation in `pd.Timestamp`.

FYI @khrapovs, and thanks for all your work on this which I feel has clarified a lot to me!

### DIFF
--- a/steps/17_forecasting_horizon_freq/step.md
+++ b/steps/17_forecasting_horizon_freq/step.md
@@ -150,3 +150,32 @@ Preferably:
 * `freq` strings as used in `pandas`
 
 The function `get_freq` can then be invoked at the start of any method with a `freq` input to convert different possible input options to a uniform `freq` representation.
+
+### 3. turn `cutoff` into a `pd.Index`
+
+Time-like `pd.Index` objects carry a `freq` attribute and will also not lose it in the above-mentioned deprecation.
+
+This motivates the following solution: turning forecasters' `cutoff` into a one-element `pd.Index` with `freq`.
+
+The solution combines the following:
+
+* the possibility to pass `freq` strings to methods of `ForecastingHorizon`, like in 2.
+* turning `_cutoff` and possibly `cutoff` attributes into  length 1 `pandas.Index` objects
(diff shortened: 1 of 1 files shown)


fkiraly:
Partial implementation here: https://github.com/alan-turing-institute/sktime/pull/2909
(`BaseForecaster` sided refactor, not covering `ForecastingHorizon`)

steps/17_forecasting_horizon_freq/step.md
```diff
@@ -149,4 +149,33 @@ Preferably:
...
+
+If utilities are extended with polymorphic signatures, downwards compatibility is also maintained. 
+This means:
+* where utilities accept index element input, they should also accept length 1 `pd.Index` input
+* where utilities return index element output, they should be able to return lenght 1 `pd.Index` input
+* which of the two outputs is returned is controlled by a `return_index` boolean argument (default=`False`)
+
+Optionally, `cutoff` in index element format can be deprecated in favour of `_cutoff`.
+
+Utilities affected by this change:
```
khrapovs, line 178: I do not see any discussion about the effect on public methods like `ForecastingHorizon.to_absolute` or `.to_relative`. These methods consume `cutoff` which is provided by a user, that is we have no control over it. We can do whatever acrobatics internally to avoid changing public interfaces, but it does not solve the problem of a user passing `pd.Timestamp` as a `cutoff` to `.to_relative`.

khrapovs (changes requested):

steps/17_forecasting_horizon_freq/step.md
fkiraly: `ForecastingHorizon` would be as in part 2, as the first bullet point says, i.e., as in your proposed solution.

In the case where the user passes `pd.Timestamp`, in the future we will also expect them to pass a `freq` string at some point before it is required, e.g., anytime in or before `to_absolute` is being called, like in your write-up.

khrapovs (approved):

PR #28, opened 2022-12-25
Update STEP on serialization of `sktime` estimators with diffs in implementation

blazingbhavneek:
@fkiraly @achieveordie Kindly check the documentation and let me know if I underwrote or overwrote. And suggest changes/advise me for future tasks.

### DIFF
--- a/steps/20_saveload/step.md
+++ b/steps/20_saveload/step.md
@@ -286,6 +286,8 @@ It contains the following files:
 
 - `_metadata` : contains class of self, i.e., type(self)
 - `_obj` : serialized self. This class uses the default serialization (pickle).
+- `/keras`: a sub directory with dumped self.model_
+- `/history`: a sub directory with dumped self.history.history
 
 
 ### Code design: estimator methods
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Great, thanks!

Can you kindly summarize the serialization file format that is valid in any case?
I.e., there are no two different file formats for deep learning and non-deep learning.

blazingbhavneek:
Hey! 
I thought this->

### zip file 
`skt` file format was proposed, `zip` was implemented
All files are stored in the `zip` format, with specification as follows.
It contains the following files:
- `_metadata` : contains class of self, i.e., type(self)
- `_obj` : serialized self. This class uses the default serialization (pickle).

summarized general file format, because the zip made from DL models do contain 2 extra sub-dirs by difference in implementation, but the file format is still same .zip, please elaborate what I need to do.

fkiraly:
> please elaborate what I need to do.

well, the old format description has a bullet point that says that other folders etc can be used for specific models. That should be contained in the main format description.
The deep learning file format is just a special case, so you should describe the general case.

blazingbhavneek:
Hey! Happy new year! I'm so sorry for the delay in a small task. @fkiraly Kindly check the latest commit and suggest changes

fkiraly (changes requested):
Thanks! Sorry, missed this.

I still think you need to describe the format that is general. Currently, e.g., in the section "zip file", the description is specific for `keras` based estimators, and incorrect for the general case.

Further, it would be great if you can avoid duplications from copy-pasting user journey etc.

fkiraly (approved):
This looks accurate, thanks! Nice!!

PR #29, opened 2023-01-28
STEP: flag manager and config interface

fkiraly:
This STEP introduces an abstract flag manager and a config interface, which parallels the tag interface.

An implementation is available in https://github.com/sktime/sktime/pull/3822

### DIFF
--- a/steps/21_config_manager/step.md
+++ b/steps/21_config_manager/step.md
@@ -94,9 +94,9 @@ config_dict = my_estimator.get_config()
 ### Code design: flag manager
 
 The flag manager is a mixin class `_FlagManager`, which provides abstract
-functionality for tags and configs.
+functionality for tags and configs. From this, `BaseObject` can inherit.
 
-The methods are private, as they are used within public methods like `get_tag`,
+`_FlagManager`'s methods are private, as they are used within public methods like `get_tag`,
 and all have an argument that provides a reference for the attribute dict
 containing the flags, e.g., `_tags` or `_config`.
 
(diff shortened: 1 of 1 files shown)


fkiraly:
FYI @RNKuhns

RNKuhns:
> FYI @RNKuhns 

@fkiraly I will take a look.

RNKuhns (changes requested):
@fkiraly I will send along a few minor suggested text tweaks. But I agree with this in theory.

However, 2 items I'd like to see addressed are: 

- Why not just add the flag manager logic to BaseObject directly, instead of having it inherit from the flag manager? Are there uses other than through BaseObject that I'm not considering (hence the mixing is useful to have around)?
- global config system (through singleton like we talked about a while back). I think for the config part a user should be able to it locally (on estimator/base object) or globally (indicate not to do input checks globallly). I'm thinking base config for should be looked up from this global config, unless overridden. I can provide some text and examples on that if it is helpful.

fkiraly:
> Why not just add the flag manager logic to BaseObject directly, instead of having it inherit from the flag manager?

That would be an option I wouldn't be opposed to, I just thought:

* that would increase the number of methods in `BaseObject` even more - "an object/function should do one thing and only one thing" and so on. Arguably that depends on the definition of what thing `BaseObject` is, of course, but I find the "flag management" concern it's better localized this way.
* I can also imagine a case where one may like to manage tags/flags in an object that is not `BaseObject descendant, although this is more of a hypothetical right now

> I'm thinking base config for should be looked up from this global config, unless overridden. I can provide some text and examples on that if it is helpful.

Yes, that would be helpful!

I haven't figured out how to avoid coupling the `BaseObject` to the global config. One idea would be an attribute that the flag manager mixin looks up, and you can hard code it in the consumer package `BaseObject`, which inherits from `skbase.BaseObject`, but that seems a little clunky.

Another question I have, should we merge this in `sktime` 0.16.0, so we can build input/output check control on it, or should we wait?

fkiraly:
(added discussion of alt design)

PR #35, opened 2024-06-12
(WIP) Design of a `BayesianEstimator` Class for `sktime` and `skpro`

meraldoantonio:
Bayesian estimators uniquely integrate prior knowledge through prior distributions and update this knowledge based on observed data to form posterior distributions. To accommodate the distinctive workflow of Bayesian estimators, we propose designing a new BayesianEstimator class.

I'm opening this PR to discuss various ideas for implementing this template. Any contributions or ideas are welcome!

### DIFF
--- a/TEMPLATE.md
+++ b/TEMPLATE.md
@@ -1,24 +1,123 @@
-# [Title]
+# Enhancement Proposal: Design of a `BayesianEstimator` Class for `sktime` and `skpro`
 
-Contributors: [list of contributors' GitHub handles]
+Contributors: [meraldoantonio]
 
 ## Introduction
-
-[concise introduction to problem and overview of proposed solution]
-
-For preliminary discussions of the proposal presented here, see issue: [links to issues/pull requests]
+This proposal outlines the design for a new `BayesianEstimator` class to be integrated into `sktime` and `skpro`. 
 
 ## Contents
 [table of contents]
 
 ## Problem statement
-[concise problem statement]
+Bayesian estimators are unique because they integrate prior knowledge through prior distributions and update this knowledge based on observed data to form posterior distributions. To accommodate the distinctive workflow of Bayesian estimators, we propose designing a new `BayesianEstimator` class. This class aims to offer robust and flexible methods for prior specification, model specification, and visualization of both prior and posterior distributions, among other unique Bayesian functionalities, while still ensuring compatibility with the existing `sktime` and `skpro` `fit`/`predict` frameworks. 
 
-## Description of proposed solution
+Initially, the focus will be on continuous tabular regression.
 
-## Motivation
+## Special Aspects and Requirements of Bayesian Estimators
 
-## Discussion and comparison of alternative solutions
+### Prior Specification
+- **Methods for Prior Specification**: The estimator should allow users to specify priors for each parameter explicitly. This can be done through methods that accept common prior distributions (e.g., Normal, Beta, Gamma) or custom user-defined distributions.
+- **Hyperparameter Tuning**: Functionality to tune hyperparameters of priors through cross-validation or other methods.
+
+### Model Specification
+- **Model Building**: Support for defining Bayesian models, including specifying likelihood functions and linking them with priors.
+- **Modular Design**: The class should be modular to support various Bayesian paradigms (e.g., Conjugate Bayesian analysis, MCMC, Variational Inference).
+
+### Posterior Inference
+- **Inference Methods**: Implement methods for obtaining posterior distributions using MCMC (e.g., Metropolis-Hastings, Gibbs Sampling) or Variational Inference.
+- **Posterior Summarization**: Methods to summarize posterior distributions (mean, median, credible intervals).
+
+### Visualization
+- **Prior and Posterior Visualization**: Methods to visualize prior and posterior distributions, including density plots, trace plots for MCMC diagnostics, and convergence diagnostics.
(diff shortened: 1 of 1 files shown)


meraldoantonio:
Linking https://github.com/sktime/skpro/pull/358 (skpro's WIP `BayesianLinearRegressor` class) as an application of this template

fkiraly:
Related: intermediate base class from `prophetverse` by @felipeangelimvieira:
https://github.com/felipeangelimvieira/prophetverse/blob/main/src/prophetverse/sktime/base.py
(imo: this is at a lower degree of generality than what we may want, but is pertinent design-wise)

fkiraly:
First high-level comments: I think we ought to write up what the key workflowelements are that every Bayesian model needs to support.

I think:

* passing prior distributions at construction
* inspecting or retrieving posterior distributions
* Bayesian update of an already fitted model

Thoughts about the state transitions and related interfaces:

* priors must be passed before `fit`. So there are only two option: in `__init__` in a dedicated argument (or multiple), or in a separate method. So, three options: init, multiple args; init, single arg; separate method
* posteriors can be retrieved anytime after `fit`. There already exists an interface point for this, `get_fitted_params`. Again a second option is adding a specialized method. Again, three options: `get_fitted_params`, multiple params; same, one param; separate method
* Bayesian update is similar to general update, such as the `update` method in forecaster. This does not exist in `skpro` tabular regressors, but could be added, similar to the `river` interface.

Further thoughts: for priors and posteriors, it might be worth using a structured object, similar to `pymc` handles them. Another question is whether we want to use `skpro` distributoins for inputs or outputs, or another standard.

Some preliminary desiderata:

* proliferation of methods should be avoided, all other things being equal
* unified interfaces that are easy to comply with are better than those that are difficult to follow, and that´s better than non-unified

From this, some slight preferences:

* mapping priors on `__init__`, and posteriors on `get_fitted_params`
* unification: single `prior` param argument and `posterior_` fitted param
* that means, we need a `BaseObject` child class for priors/posteriors.
    * this could be a `BaseDistribution` representing multiple, potentially associated random variables?

fkiraly:
FYI @felipeangelimvieira

meraldoantonio:
I've updated the design document to include these ideas, and I believe the desiderata make sense. Mapping priors in `__init__` and posteriors in `get_fitted_params` is also logical.

However, I have two areas where I'm unsure of:

**Representing Multiple Associated Random Variables**:
Can we use a single BaseDistribution class to represent multiple, potentially associated random variables? This approach would need careful consideration to handle dependencies and interactions between the variables effectively.

**Bayesian Workflow Integration**:
Besides passing a list of priors in `__init__`, we need to include information on how these priors build on each other to form a Bayesian workflow. Should this be part of `__init__`, or do we "freeze" the workflow within a class, making each class specific to a workflow (e.g., Bayesian regression as one class)?

felipeangelimvieira:
Hi @meraldoantonio! 
Recently started reading more about PyMC, and they have just added a custom `Prior` class in their "PyMC-Marketing" lib. Maybe this can help with ideas: [code here](https://github.com/pymc-labs/pymc-marketing/blob/5cda250e53ae80d42eef2e60c4eb3c23ae1c3c0e/pymc_marketing/prior.py#L219
)

meraldoantonio:
Re-opening this thread and this work stream.

Following @felipeangelimvieira's suggestion - I looked at `pymc-marketing`'s `Prior` class and it looks like a suitable way to specify the prior distributions. It is flexible and is native to `pymc` ecosystem. Also, the fact that the library already comes with the [Bayesian template ("ModelBuilder")](https://github.com/pymc-labs/pymc-marketing/blob/5cda250e53ae80d42eef2e60c4eb3c23ae1c3c0e/pymc_marketing/model_builder.py) means that we can repurpose a lot of functionalities for our purpose.
An example of the application of the ModelBuilder is [here](https://github.com/pymc-labs/pymc-marketing/blob/35c8850811b2467ad1797ce0bb0e9b5db06606e2/pymc_marketing/mmm/base.py#L52).

meraldoantonio:
Copying @fkiraly 's points from our meeting:
* reflections:
    * is there option to "mix"? E.g., have a "simple" option with prescribed priors and configurable?
    * are there any dependencies that are specific to the specification? E.g., if we use `Prior` from `pymc-marketing`, would that imply more deps than the models have?
        * FK: if estimator also uses `pymc-marketing` internally, that's probably ok
        * but not all Bayesian estimators will use `pymc-marketing`. I think taking a dependency for pure speciication reasons is a bad idea
            * story/experience: originally we wanted to rely on `tensorflow-probability` for the proba distributions, as return objects; but this turned out a bad choice since `tensorflow` got less popular and the proba distirbutions got less maintained, yet implied a heavy dependency footprint for the sole reason of specification
* minor points:
    * what is the reliability of `pymc-marketing` - usage and maintenance footprint, this is not pymc itself

meraldoantonio:
Summary of updates:

- **Enhancement Proposals Update:**  
Provided workflow ideas for the Bayesian class and included an example implementation (`BayesianLinearRegression`). This class utilizes `PyMC` for inference, `ArViz` for data handling, and `pymc-marketing`'s `Prior` for prior specification. The updated `Template.md` justifies these design choices.

- **Implementation Update:**  
  Enhanced the `BayesianLinearRegression` class in [skpro](https://github.com/sktime/skpro/pull/358). Included an updated implementation and a notebook demonstrating its usage.

PR #38, opened 2024-12-30
K-visibility: new clustering and classification algorithm 

seigpe:


fkiraly:
Thanks!

A STEP is not necessary for adding a new algorithm, I will just close this in favour of the issue on sktime.

PR #39, opened 2025-01-31
`pytorch-forecasting` and `dsipts` v2 API design

fkiraly:
An enhancement proposal and design document towards v2 of `pytorch-forecasting` and `dsipts`.

More context here: https://github.com/sktime/pytorch-forecasting/issues/1736

### DIFF
--- a/steps/25_pytorch_forecasting_dsipts/step.md
+++ b/steps/25_pytorch_forecasting_dsipts/step.md
@@ -53,12 +53,62 @@ References:
 Our proposed solution consists of the following components:
 
 * a two-layered `DataSet` input layer: first layer generic input, second layer. The first layer (layer D1) is unified and model-independent, the second layer (layer D2) is model or model class specific.
-* a two-layered model layer: a composite layer. The inner layer is pure `torch` (layer T), the outer layer (layer C) is a composite of metadata, references to layer D2, and to layer T.
+* a two-layered model layer: a composite layer. The inner layer is pure `torch` (layer T), the outer layer (layer M) provides a unified interface, is a composite of metadata, references to layer D2, and to layer T.
 * downwards compatible migration and refactoring strategies for `pytorch-forecasting`, `dsipts`, and `thuml`, towards a unified whole that also leaves current structures intact.
 
-## Design: foo
+## Design: `pytorch-forecasting` 2.0
 
-### Conceptual model: bar
+### Conceptual model and layers
+
+Following discussions collected in the linked issue [1736](https://github.com/sktime/pytorch-forecasting/issues/1736),
+
+the design consists of four layers as mentioned above:
+
+* layer D1: unified `DataSet` interface
+* layer D2: model specific `DataSet` and `DataLoader` interface
+* layer T: raw `torch` models
+* layer M: unified model layer: models with metadata and reference to d2 layer
+
+Reasoning for the layers:
+
+* both `DataSet` and `torch` model input/output tend to be specific to implementation.
+From examples seen, it is unlikely that they can be sensibly unified.
+* therefore, additional unification layers are needed, one for data and one for models.
+* discussions in the issue (also motivated by notes of janbeitner in the original code)
+converged on two `DataSet` based layers, using the standard `DataLoader`.
+* this implies a second unification layer on the other side, for models - given that
+unifying model interfaces is the primary goal.
+
+Conceptually, layers align with concepts as follows:
+
+* layer D1: abstract data type of "collection of time series", a `Panel` in `sktime`
+parlance. Implementation
+can be arbitrary, e.g., `pandas` or `polars`, or hard-drive files.
+* layer M: abstract model taking in the `Panel` data for training or inference.
+This *includes* data pre-processing, re-sampling, batching.
+* layer T: concrete neural network with free parameters, *excluding* data pre-processing, re-sampling, batching.
+This starts at data that is already pre-processed, re-sampled, batched.
+* layer D2: M minus T.
+
+### Alignment of current packages with layers
+
+`pytorch-forecasting`:
+
+* currently has two layers, a data layer and a model layer
+* data layer = D1 plus D2 plus M (lasagna) = `TimeSeriesDataSet`
+* model layer = T
+* in particular, there is no uniformization layer for data or models that would cover, e.g., foundation models
+* this also makes the design of very limited extensibility beyond certain decoder/encoder models
+
+`dsip-ts`:
+
+* currently has three layers
+* pre-processing functions, prior to use of `DataSet` - D1
+* data set and data loader: D2 plus M
+* model layer = T
+* improvement compared to `pytorch-forecasting`, because there is a data uniformization layer
+    * but unfortunately D1 is not in the form of `DataSet` which would allow scaling
+    * and there is no model uniformization layer
 
 
 ### foo
--- a/steps/25_pytorch_forecasting_dsipts/step.md
+++ b/steps/25_pytorch_forecasting_dsipts/step.md
@@ -60,18 +60,6 @@ Our proposed solution consists of the following components:
 
 ### Conceptual model: bar
 
-Conceptually, we want to model "distribution objects" that are probability distributions, taking values in `pandas.DataFrame`-s, possibly with hierarchical row index.
-
-For this, we need to keep track of the following information:
-
-* a batch/instance structure of the distribution, e.g., 2D matrix shape
-* distribution defining information, i.e., parameters of the distribution
-* row index and column index of the distribution, both `pd.Index`.
-
-Row and column index shape need to agree with the batch/instance shape of the distribution.
-
-Typically but not nececssarily, different rows will be independent, but different columns in the same row will not be independent.
-
 
 ### foo
 
--- a/steps/25_pytorch_forecasting_dsipts/step.md
+++ b/steps/25_pytorch_forecasting_dsipts/step.md
@@ -97,6 +97,7 @@ This starts at data that is already pre-processed, re-sampled, batched.
 * currently has two layers, a data layer and a model layer
 * data layer = D1 plus D2 plus M (lasagna) = `TimeSeriesDataSet`
 * model layer = T
+* `BaseModel` is similar to C, but assumes data layer
 * in particular, there is no uniformization layer for data or models that would cover, e.g., foundation models
 * this also makes the design of very limited extensibility beyond certain decoder/encoder models
 
@@ -111,12 +112,247 @@ This starts at data that is already pre-processed, re-sampled, batched.
     * and there is no model uniformization layer
 
 
-### foo
+### mid-level interfaces
+
+#### layer D1
+
+Aim: model `Panel` data as closely as possible, while satisfying data requirements
+
+Data requirements:
+
+* agnostic towards data location - `pandas`, `polars`, hard drive
+* capturing metadata: numeric/categorical, past/future known, dynamic/static
+
+Design:
+
+* `DataSet` extension API with unified `__getitem__` output, defined by `BaseTSDataSet`
+* `__init__` captures input that can vary
+    * for downwards capability, current inputs in `pytorch-forecasting` and `dsipts` supported
+* inheritance pattern and strategy pattern
+* simplest-as-possible `__getitem__` return
+
+##### interface: proposed `__getitem__` return of `BaseTSDataSet`
+
+As implemented in draft [PR 1757](https://github.com/sktime/pytorch-forecasting/pull/1757)
+
+```
+    Sampling via ``__getitem__`` returns a dictionary,
+    which always has following str-keyed entries:
+    * t: tensor of shape (n_timepoints)
+      Time index for each time point in the past or present. Aligned with ``y``,
+      and ``x`` not ending in ``f``.
+    * y: tensor of shape (n_timepoints, n_targets)
+      Target values for each time point. Rows are time points, aligned with ``t``.
+      Columns are targets, aligned with ``col_t``.
+    * x: tensor of shape (n_timepoints, n_features)
+      Features for each time point. Rows are time points, aligned with ``t``.
+    * group: tensor of shape (n_groups)
+      Group ids for time series instance.
+    * st: tensor of shape (n_static_features)
+      Static features.
+    * y_cols: list of str of length (n_targets)
+      Names of columns of ``y``, in same order as columns in ``y``.
+    * x_cols: list of str of length (n_features)
+      Names of columns of ``x``, in same order as columns in ``x``.
+    * st_cols: list of str of length (n_static_features)
+      Names of entries of ``st``, in same order as entries in ``st``.
+    * y_types: list of str of length (n_targets)
+      Types of columns of ``y``, in same order as columns in ``y``.
+      Types can be "c" for categorical, "n" for numerical.
+    * x_types: list of str of length (n_features)
+      Types of columns of ``x``, in same order as columns in ``x``.
+      Types can be "c" for categorical, "n" for numerical.
+    * st_types: list of str of length (n_static_features)
+      Types of entries of ``st``, in same order as entries in ``st``.
+    * x_k: list of int of length (n_features)
+      Whether the feature is known in the future, encoded by 0 or 1,
+      in same order as columns in ``x``.
+      0 means the feature is not known in the future, 1 means it is known.
+    Optionally, the following str-keyed entries can be included:
+    * t_f: tensor of shape (n_timepoints_future)
+      Time index for each time point in the future.
+      Aligned with ``x_f``.
+    * x_f: tensor of shape (n_timepoints_future, n_features)
+      Known features for each time point in the future.
+      Rows are time points, aligned with ``t_f``.
+    * weight: tensor of shape (n_timepoints), only if weight is not None
+    * weight_f: tensor of shape (n_timepoints_future), only if weight is not None
+```
+
+##### Extension pattern
+
+* inherit from `BaseTSDataSet`
+* custom `__init__` input, can be anything, including file locations
+* dataclass-like
+* logic only needs to comply with `__getitem__` expectation
+
+
+#### layer D2
+
+Aim: prepare unified data input from layer D1 for `torch` model
+
+Design:
+
+* `DataSet` extension API with unified `__init__` input, expecting `BaseTSDataSet`
+* further `__init__` fields may be arbitrarily present, dataclass-like
+* `__getitem__` return is specific to a limited range of `torch` models
+* default assumption is standard `DataLoader`
+* optionally, custom `DataLoader` may be supplied
+
+##### Example, based on current `pytorch-forecasting` models
+
+Current `TimeSeriesDataSet(data, **params)` to be replaced with
+
+```python
+tsd = PandasTSDataSet(df, **metadata)  # layer D1
+DecoderEncoderData(tsd, **params_without_metadata)  # layer D2
+```
+
+* where `metadata` is as above in layer D1
+* and `params_without_metadata` contains decoder/encoder specific variables
+    * `max_encoder_length`
+    * `min_encoder_length`
+    * `max_decoder_length`
+    * `min_decoder_length`
+    * `constant_fill_strategy`
+    * `allow_missing_timesteps`
+    * `lags`
+    * and so on
+
+The return of the `DecoderEncoderData` instance` should be exactly the same
+as of current `TimeSeriesDataSet`, when invoked with equivalent parameters and data.
+
+##### Example, based on current `dsip-ts` models
+
+For custom data, this should work
+
+```python
+tsd = PandasTSDataSet(df, **metadata)  # layer D1
+DsiptsPipeline(tsd, **params_without_metadata)  # layer D2
+```
+
+For pre-defined datasets, this should work
+
+```python
+tsd = BenchmarkDataSet(name:str, config)  # layer D1
+DsiptsPipeline(tsd, **params_without_metadata)  # layer D2
+```
+
+#### layer T
+
+The model layer contains layers and full models using `pytorch-lightning` interfaces.
+
+These are simple loose classes as currently present in all packages, i.e., `nn.Module` subclasses.
+
+
+#### layer M
+
+Some unknowns here and work in progress.
+
+Suggested design:
+
+* class design: metadata class with pointer to layer T and D2, plus metadata
+* `scikit-base` compatible collection of parameters
+    * neural network parameters (`torch.nn`)
+    * training and inference parameters
+* switch between training and inference mode
+* directly interfaces with layer D1 on the outside
+    * possibility to construct `from_dataset` or similar, like in ptf
+
+
+```python
+class MyNetwork(BasePtfNetwork):
+
+    _tags = {
+        "capability:categorical": True,
+        "capability:futureknown": False,
+        "capability:static": False,
+        "etc"
+    }
+
+    def __init__(
+        self,
+        **network_params,
+        **network_configs,
+        **loader_params,
+    )
+
+    def ref_network(self):  # pointer to network, could be more complicated
+        from somewhere import MyTorchNetwork
+
+        return MyTorchNetwork
+
+    def ref_dataloader(self):  # pointer to dataloader
+        from somewhere import D2LoaderForMyTorchnetwork
+
+        return D2LoaderForMyTorchnetwork
+
+    @classmethod
+    def from_dataset(cls, dataset):  # sets parameters from dataset
+        return cls(**get_params_from(dataset))
+
+    def should_we_forward_lignthing_methods(self, **kwargs):  # ?
+
+    def train(self, dataset):
+        # logic related to training
+
+    def predict(self, dataset)
+        # logic related to inference
+```
+
 
 ## Change and deprecation
 
-### Change of foo
+### `pytorch-forecasting`
+
+* Networks can be left as-is mostly, for downwards compatibility
+
+* `TimeSeriesDataSet` should alias this, see above
+
+```python
+tsd = PandasTSDataSet(df, **metadata)  # layer D1
+DecoderEncoderData(tsd, **params_without_metadata)  # layer D2
+```
+
+It should be possible to keep interfaces as-is with this aliasing.
+
+### `dsip-ts`
+
+* need to introduce a D1-to-D2 `DataSet`
+* current pipeline can still be used
 
 ## Implementation phases
 
-### Phase 1 - bar
+### Phase 0 - design
+
+Agreement on this design document and target state
+
+### Phase 1 - `DataSet` layer
+
+Suggested to use `pytorch-forecasting` and introduce D1/D2 separation as an API
+preserving refactor of `TimeSeriesDataSet`, as follows:
+
+1. add D1 `BaseTSDataSet` and `PandasTSDataSet` child class, and tests
+2. add `DecoderEncoderData` to obtain interface on par with `TimeSeriesDataSet`
+3. change `TimeSeriesDataSet` to alias the D1/D2 composite
+4. add one or two further `BaseTSDataSet` as proof-of-concept: `polars` or hard drive files
+    * use this to improve `DecoderEncoderData` to avoid too high in-memory usage
+
+
+### Phase 2a - `dsipts` `DataSet` layer integration
+
+Can start in middle of phase 1, at a stage where `BaseTsDataSet` is consolidated.
+
+1. refactor current data pipeline to be a single `DataSet` class.
+2. rebase pipeline on `BaseTSDataSet` interface, ensure refactor and API consistency
+
+### Phase 2b - Model layer
+
+1. `BasePtfNetwork` experimental design and full API tests, using phase 1 objects
+2. refactor at least two `pytorch-forecasting` models to this design, design iteration
+
+### Phase 3 - ecosystem
+
+* `dsip-ts` models
+* `pytorch-forecasting` models
+* `thuml` models
--- a/steps/25_pytorch_forecasting_dsipts/step.md
+++ b/steps/25_pytorch_forecasting_dsipts/step.md
@@ -1,0 +1,84 @@
+# API rework of pytorch-forecasting and dsipts merge (pytorch-forecasting version 2)
+
+Contributors: @fkiraly, @agobbifbk
+
+## High-level summary 
+
+### The Aim
+
+To create a unified interface for `torch` forecasters, in `pytorch-forecasting` version 2,
+suitable for the entire ecosystem of `torch` based forecasters, including `dsip-ts`,
+various foundation models, and inspired by `dsip-ts` and `time-series-library`.
+
+### Context
+
+Over 2024, in `sktime`, interfaces to a variety of deep learning and foundation models have been added,
+amongst these, `pytorch-forecasting`, and various foundation models.
+
+This exercise showed that the `pytorch-forecasting` design does not generalize to foundation models,
+and made some limitations of the package apparent, such as strong coupling to `pandas` and `scikit-learn`
+which prevents large-scale use.
+
+At the same time, the `dsip-ts` package (by agobbifbk) emerged, contributing interesting ideas
+of API uniformity, and a simple API.
+
+It was decided that both packages - `pytorch-forecasting` and `dsip-ts` - would merge
+with the aim to creat `pytorch-forecasting` - as the "sktime" of `torch` forecasting models.
+
+References:
+
+* `pytorch-forecasting` handover plan (to `sktime`) https://github.com/sktime/pytorch-forecasting/issues/1592
+* re-design thread for `pytorch-forecasting` 2.0 with `dsip-ts` https://github.com/sktime/pytorch-forecasting/issues/1736
+    * this thread also contains a summary of technical planning meetings and sync design discussions
+* umbrella issue `sktime` on foundation models https://github.com/sktime/sktime/issues/6177
+* early coordination discussion deep learning, foundation models, `pytorch-forecasting` https://github.com/sktime/sktime/issues/6381
+* 2025 `sktime` roadmap with a focus on deep learning and `torch` https://github.com/sktime/sktime/issues/7707
+
+### requirements
+
+* M: unified model API which is easily extensible and composable, similar to `sktime` and DSIPTS, but as closely to the `pytorch` level as possible. The API need not cover forecasters in general, only `torch` based forecasters.
+    * M: unified monitoring and logging API, also see https://github.com/sktime/pytorch-forecasting/issues/1700
+    * M: extension templates need to be created
+    * S: `skbase` can be used to curate the forecasters as records, with tags, etc
+    * S: model persistence
+    * C: third party extension patterns, so new models can "live" in other repositories or packages, for instance `thuml`
+* M: reworked and unified data input API
+    * M: support static variables and categoricals
+    * S: support for multiple data input locations and formats - pandas, polars, hard drive, distributed, etc
+* M: MLops and benchmarking features as in DSIPTS
+* S: support for pre-training, model hubs, foundation models, but this could be post-2.0
+
+### The proposed solution
+
+Our proposed solution consists of the following components:
+
+* a two-layered `DataSet` input layer: first layer generic input, second layer. The first layer (layer D1) is unified and model-independent, the second layer (layer D2) is model or model class specific.
+* a two-layered model layer: a composite layer. The inner layer is pure `torch` (layer T), the outer layer (layer C) is a composite of metadata, references to layer D2, and to layer T.
+* downwards compatible migration and refactoring strategies for `pytorch-forecasting`, `dsipts`, and `thuml`, towards a unified whole that also leaves current structures intact.
+
+## Design: foo
+
+### Conceptual model: bar
+
+Conceptually, we want to model "distribution objects" that are probability distributions, taking values in `pandas.DataFrame`-s, possibly with hierarchical row index.
+
+For this, we need to keep track of the following information:
+
+* a batch/instance structure of the distribution, e.g., 2D matrix shape
+* distribution defining information, i.e., parameters of the distribution
+* row index and column index of the distribution, both `pd.Index`.
+
+Row and column index shape need to agree with the batch/instance shape of the distribution.
+
+Typically but not nececssarily, different rows will be independent, but different columns in the same row will not be independent.
+
+
+### foo
+
+## Change and deprecation
+
+### Change of foo
+
+## Implementation phases
+
+### Phase 1 - bar


felipeangelimvieira:
(bringing the discussion from discord to here)
Really interesting, Franz... I’ll post a more detailed comment here once I’ve organized my thoughts. The first question that comes to mind is: could we condense the information in the output of `__getitem__` of BaseTSDataset into fewer objects?

felipeangelimvieira:
About `static_features`: not all deep-learning TS models have static features. Example: N-BEATS. At the same time, I believe this is something really relevant to TFT, for example.

Given that, should this be positioned at "model" level instead?

fkiraly:
>The first question that comes to mind is: could we condense the information in the output of `__getitem__` of BaseTSDataset into fewer objects?

Yes, that is also what @thawn said, let's try to have as few tensors as possible.

My answer: perhaps - but it is hard to actually write it down.

Some comments:

- a lot of the fields are metadata.
- we need at least 2 if we have only past data, and 3 if we have future data, simply given the native "shapes" of static, future and past data. Tensors can be rectangular only, and it is not possible to have all that data in less than 3 because of the shapes.
- separating index is a conscious choice, otherwise you have an additional metadata field. And you are open to a future modification where you encode a `RangeIndex` or period index more efficiently.

fkiraly:
> About `static_features`: not all deep-learning TS models have static features. Example: N-BEATS. At the same time, I believe this is something really relevant to TFT, for example.

> Given that, should this be positioned at "model" level instead?

I think not, since it is a "data" concern:

* The information on which variable is static is a data property
* the "minimal representation" of static variables involves a tensor of smaller shape (n_instances x n_static_variables) - 2D and not 3D (instance, variable, time)
* the corresponding "model" concern is "handles static variables", this is a capability type property. For a unified interface, all models need to be able to take a dataset with static variables, the ones without capability will just ignore them.

felipeangelimvieira:
What I'm thinking about is that static variables are a specific kind of past and future known exogenous variables, that are constant throughout time. In this sense, should we really separate them? I feel they are essentially the same in a "meta" level. The difference is how the model wants to use them.

One example: we could use fourier features as static variables in TFT. I'm not sure if there's something that do not allow us to use it that way, since these features are known for every timepoint in past and future.

felipeangelimvieira:
More thoughts:

- Couldn't `__getitem__` return just the relevant data, and metadata be stored in attributes `self.y_cols_`, `self.x_cols_` of the dataset?
- Should we have a `BaseTorchTransformer` that defines an interface for transformations (such as normalization of past values in a batch) at torch level? (replacing [current encoders](https://pytorch-forecasting.readthedocs.io/en/stable/api/pytorch_forecasting.data.encoders.html))

fkiraly:
> What I'm thinking about is that static variables are a specific kind of past and future known exogenous variables, that are constant throughout time. In this sense, should we really separate them? I feel they are essentially the same in a "meta" level. The difference is how the model wants to use them.

To me, this is not correct, on the conceptual level.

The formal model I typically adopt is that all data points are indexed. Time-varying (or "dynamic") variables change over time, and obervations (entries, values, etc) are indexed by: (instance index, variable name, time index). Whereas, static variables are only indexed by (instance index, variable name). If you adopt the usual model of hierarchical data, where a combination of indices indexes an observation, it becomes clear that static variables live at different "levels" in the indexing hierarchy.

Mapping static variables on a column on the lower level (plus time index) has them being constant as a side effect in representation, rather than as a conceptual feature.

fkiraly:
> - Couldn't `__getitem__` return just the relevant data, and metadata be stored in attributes `self.y_cols_`, `self.x_cols_` of the dataset?

There are two ideas here, I believe: (a) nesting metadata, and (b) returning metadata not by `__getitem__` but another API point.

I am very open to (a) and think this is perhaps a good idea. Regarding (b), could you explain how the retrieval of the metadata would work? The `__getitem__` return seems to be the single intended interface point, if you go through the entire `lightning` workflow, e.g., `DataLoader` does not assume anything except (if I understand correctly). If you disagree, I would appreciate an explanation how the information in the suggested attributes will reach the model, if it needs to.

> Should we have a `BaseTorchTransformer` that defines an interface for transformations (such as normalization of past values in a batch) at torch level? (replacing [current encoders](https://pytorch-forecasting.readthedocs.io/en/stable/api/pytorch_forecasting.data.encoders.html))

Yes, that would be a great idea - do you want to draw up a design? And, surely this exists elsewhere, too?

felipeangelimvieira:
> The formal model I typically adopt is that all data points are indexed. Time-varying (or "dynamic") variables change over time, and obervations (entries, values, etc) are indexed by: (instance index, variable name, time index). Whereas, static variables are only indexed by (instance index, variable name). If you adopt the usual model of hierarchical data, where a combination of indices indexes an observation, it becomes clear that static variables live at different "levels" in the indexing hierarchy.
> 


I understand your point that, since they can be indexed by different amout of indexes while keeping the information, they are and can be classified as different things. Nevertheless, I believe that the challenge here is projecting this higher dimensional space of concepts (e.g. classifying variables by instance, variable name, time index, discrete or continuous, etc) into a lower dimensional space that still makes the task (implementing different NN-based forecasters) feasible.

In this regard, the separation of data in 1. target variable, 2. features with past knowns, 3. features with past/future knowns is possibly the lowest dimensional representation, and that can be further subdivided into other categories/concepts, such as features  with past/future knowns that are static and do not change overtime (we can imagine it as a tree).

> Mapping static variables on a column on the lower level (plus time index) has them being constant as a side effect in representation, rather than as a conceptual feature.

I thinking in the opposite direction: static variables are features, they belong to this "feature" concept: relevant data other than the target variable used as input to predict the target variable. They are a special subset of features, that do not change over time, but this do not categorizes them outside the set of features. In particular, being part of this subset allows them to be represented with less index levels as you mentioned.

Some models, such as DeepAR, may still want to use "static variables" as exogenous (I believe this is the case, by inspecting the implementation and the property `self.reals` inherited from parent`BaseModelWithCovariates`). So the differentiation of static and non-static is actually model dependent. TFT, on the other hand, uses static variables in a particular way.

felipeangelimvieira:
>  am very open to (a) and think this is perhaps a good idea. Regarding (b), could you explain how the retrieval of the metadata would work? The __getitem__ return seems to be the single intended interface point, if you go through the entire lightning workflow, e.g., DataLoader does not assume anything except (if I understand correctly). If you disagree, I would appreciate an explanation how the information in the suggested attributes will reach the model, if it needs to.

good point... If this is a direction we would like to go, I believe there would be possibly to solutions:
- Create a custom dataloader
- Create a custom Tensor (inherting from pytorch's tensor, that keeps the metadata of the features).

fkiraly:
> Some models, such as DeepAR, may still want to use "static variables" as exogenous (I believe this is the case, by inspecting the implementation and the property self.reals inherited from parentBaseModelWithCovariates). So the differentiation of static and non-static is actually model dependent. TFT, on the other hand, uses static variables in a particular way.

But, this is not an intrinsic property of the static varibales. It is *how the model chooses to use them*. So it is not a data property but a modelling choice, and specifically a "reduction" choice where we reformat the data (like in the reduction forecaster, just a different type of making the data "rectangular-formatted")

Therefore, with the domain driven design philosophy, the choice of reformatting the static variables falls with the model rather than the data layer.

> Create a custom dataloader

I wanted to do that too but was voted down and convinced by good arguments in the Jan 20 meeting, reference: https://github.com/sktime/pytorch-forecasting/issues/1736

The key argument for me was, current vignettes in all packages reviewed rely on the default `DataLoader`, and details seem unclear on how this would interact with model and `lightning` layer. Naive approaches seem to lead to coupling throughout.

> Create a custom Tensor (inherting from pytorch's tensor, that keeps the metadata of the features).

I think that is a hypothetical possibility but a bad idea. We would be rewriting layers outside the scope of the package.

I still have nightmares from our attempts to rewrite `DataFrame` for `sktime`.
https://github.com/sktime/data-container

@prockenschaub did great work, and I think our designs would have been successful in the end - it just turned out we were getting pulled into the scope of rewriting `pandas` from scratch, with more resources required than we had (focusing on the core package and all).

The lesson from that is, I suppose, if you are building a house and suddenly find yourself wanting to tear down your neighbour's and then the entire street, there is maybe something architectural to rethink.

phoeenniixx:
Hi Everyone! I have an idea for D1 layer, not sure if this will work but I thought to reduce the metadata by using dicts. 
I would really appreciate the comments 

D1 `__getitem__` returns:
```
* t: tensor of shape (n_timepoints)
  Time index for each time point in the past or present. Aligned with ``y``, and ``x`` not ending in ``f``.
* y: tensor of shape (n_timepoints, n_targets)
  Target values for each time point. Rows are time points, aligned with ``t``.
* x: tensor of shape (n_timepoints, n_features)
  Features for each time point. Rows are time points, aligned with ``t``.
* group: tensor of shape (n_groups)
  Group identifiers for time series instances.
* st: tensor of shape (n_static_features)
  Static features.
* col_map: dict { 'y': list[str], 'x': list[str], 'st': list[str] }
  Names of columns for y, x, and static features.
* st_col_type: dict[str, int]
  Names of static features mapped to type (0=categorical, 1=numerical).
  Example: { 'region': 0, 'temp': 1 }
* y_types: list of int (length = n_targets)
  Type of each target variable: 0 (categorical), 1 (numerical).
* x_types: list of int (length = n_features)
  Type of each feature:  
  - 0 = categorical unknown  
  - 1 = categorical known  
  - 2 = numerical unknown  
  - 3 = numerical known
Optionally, the following str-keyed entries can be included:
    * t_f: tensor of shape (n_timepoints_future)
      Time index for each time point in the future.
      Aligned with ``x_f``.
    * x_f: tensor of shape (n_timepoints_future, n_features)
      Known features for each time point in the future.
      Rows are time points, aligned with ``t_f``.
    * weights: tensor of shape (n_timepoints), only if weight is not None
    * weight_f: tensor of shape (n_timepoints_future), only if weight is not None
```

phoeenniixx:
>  There are two ideas here, I believe: (a) nesting metadata, and (b) returning metadata not by `__getitem__` but another API point.

As we are using just `lightning` modules (please correct me if I am wrong) we can get the metadata in `setup` function which is a part of the module: https://lightning.ai/docs/pytorch/LTS/common/lightning_module.html#setup

example:
```python
class MyModel(pl.LightningModule):
    def setup(self, stage=None):
        dataset = self.trainer.datamodule.train_dataset
        self.col_map = dataset.col_map
        self.st_col_type = dataset.st_col_type
        self.y_types = dataset.y_types
        self.x_types = dataset.x_types

    def forward(self, x):
     # Now we have metadata available for processing for forward
        pass

    def training_step(self, batch, batch_idx):
    # training logic
        pass

```

image
Source: https://www.assemblyai.com/blog/pytorch-lightning-for-dummies/

More examples: https://lightning.ai/docs/pytorch/stable/data/datamodule.html#what-is-a-datamodule


Or we can do this:
```python
class DecoderEncoderData(Dataset):
    def __init__(self, tsd: PandasTSDataSet, **params):
        self.tsd = tsd  # Store dataset reference

        # Access metadata from dataset (D1)
        self.col_map = tsd.col_map
        self.st_col_type = tsd.st_col_type
        self.y_types = tsd.y_types
        self.x_types = tsd.x_types

    def __getitem__(self, idx):
        sample = self.tsd[idx]  
        return sample


```

phoeenniixx:
A doubt: are we using `pytorch-lightning` or `lightning`? As this is the thing that confuses me everytime I work on ptf, both are similar yet different and although they have similar names and functions they can't  be used interchangeably as even their `trainers` are different (although both have `setup` support) 
I am just asking this for future refrences :)

felipeangelimvieira:
> I think that is a hypothetical possibility but a bad idea. We would be rewriting layers outside the scope of the package.
What I meant is not using this tensor in every part of the code. Just having a code encapsulating the input data tensor with this metadata. This something documented in [pytorch's documentation](https://pytorch.org/docs/stable/notes/extending.html#extending-torch-python-api)

> But, this is not an intrinsic property of the static varibales. It is how the model chooses to use them. So it is not a data property but a modelling choice, and specifically a "reduction" choice where we reformat the data (like in the reduction forecaster, just a different type of making the data "rectangular-formatted")
> 
> Therefore, with the domain driven design philosophy, the choice of reformatting the static variables falls with the model rather than the data layer.

I would argue that "static" features are not a real thing, they are a separation made in the original paper but nothing impedes the usage of other past and future known features in the way "static" is used.

One example of use-case that might have troubles when separating static variables and non-static in data layer: what if I want to use non-constant features (such as fourier terms, that are known for the past and future) in the same way that static variables are used in TFT? This approach could be interesting if we wan't to have the gating mechanisms being dependent on the seasonality. As an example, holidays have different effects depending on the day of the week. The sines and cosines as static feature could be helpful in this scenario.

phoeenniixx:
Thanks for clarifying this! @felipeangelimvieira

fkiraly:
@phoeenniixx, thanks a lot for your useful thoughts!

> A doubt: are we using `pytorch-lightning` or `lightning`?

I think the latter is a rebrand of the former, the package and hte project have been renamed.

> we can get the metadata in `setup` function

Interesting - how would the entire workflow look like if we do this?

Though, using `setup` py seems to deviate slightly from its intended specification, the docs refer to data processing related setup rather than metadata handling.

> I thought to reduce the metadata by using dicts.
I would really appreciate the comments

That's a very interesting idea!

Following up on this, I would actually put even more in single dicts, and use strings (like in pandas) for types:

D1 `__getitem__` returns:
```
* t: tensor of shape (n_timepoints)
  Time index for each time point in the past or present. Aligned with ``y``, and ``x`` not ending in ``f``.
* y: tensor of shape (n_timepoints, n_targets)
  Target values for each time point. Rows are time points, aligned with ``t``.
* x: tensor of shape (n_timepoints, n_features)
  Features for each time point. Rows are time points, aligned with ``t``.
* group: tensor of shape (n_groups)
  Group identifiers for time series instances.
* st: tensor of shape (n_static_features)
  Static features.
* cols: dict { 'y': list[str], 'x': list[str], 'st': list[str] }
  Names of columns for y, x, and static features. List elements are in same order as column dimensions.
  Columns not appearing are assumed to be named x0, x1, etc, y0, y1, etc, st0, st1, etc.
* col_type: dict[str, str]
  maps column names to data types "F" (numerical) and "C" (categorical).
  Column names not occurring are assumed "F".
* col_known: dict[str, str]
  maps column names to "K" (future known) or "U" (future unknown).
  Column names not occurring are assumed "K".
  
Optionally, the following str-keyed entries can be included:
    * t_f: tensor of shape (n_timepoints_future)
      Time index for each time point in the future.
      Aligned with ``x_f``.
    * x_f: tensor of shape (n_timepoints_future, n_features)
      Known features for each time point in the future.
      Rows are time points, aligned with ``t_f``.
    * weights: tensor of shape (n_timepoints), only if weight is not None
    * weight_f: tensor of shape (n_timepoints_future), only if weight is not None
```

phoeenniixx:
Thanks for the comments @fkiraly !


> how would the entire workflow look like if we do this?

Well I have not thought this through but we can do something like this:
```python
# Layer D1
class PandasTSDataSet:
    def __init__(self, df, y_cols, x_cols, st_cols, y_types, x_types, st_types):
        self.df = df
        self.metadata = [metadata]    
    def __getitem__(self, idx):
        # Returns only tensors, no metadata
        return {
            't': tensor(...),
            'y': tensor(...),
            'x': tensor(...),
            'group': tensor(...),
            'st': tensor(...),
            't_f': tensor(...),
            'x_f': tensor(...),
        }
    
    def get_metadata(self):
        return self.metadata

# Layer D2
class DecoderEncoderDataModule(LightningDataModule):
    def __init__(self, d1_dataset, batch_size=32, num_workers=4):
        super().__init__()
        # initialize other  params
        self.d1_dataset = d1_dataset
        self.batch_size = batch_size
        self.metadata = None  # Will be set in setup()
    
    def setup(self, stage=None):
        # Get metadata from D1 layer during setup
        self.metadata = self.d1_dataset.get_metadata()
        
        if stage == 'fit' or stage is None:
            # Any train-specific setup
            pass
        
        if stage == 'test' or stage is None:
            # Any test-specific setup
            pass
            
    def train_dataloader(self):
        return DataLoader(
            self.d1_dataset,
            batch_size=self.batch_size,
            **other_params
        )

# Layer M
class DecoderEncoderModel(LightningModule):
    def __init__(self):
        super().__init__()
        self.metadata = None
        
    def setup(self, stage=None):
        # Get metadata from datamodule during setup
        self.metadata = self.trainer.datamodule.metadata
        
        # Initialize layer T model using metadata
    
    def forward(self, x):
     # forward logic
        pass

```


> Though, using `setup` py seems to deviate slightly from its intended specification, the docs refer to data processing related setup rather than metadata handling.

Well yes, but I found this was the closest way where data and related fields are handled in  the model. Or as I mentioned earlier, we can just pass the metadata directly to other layers by doing something similar to this.

Here we can even define a function like `get_metadata()` that returns everything that's necessary, that will reduce the scope of missing any metadata field.Like:

```python
class DecoderEncoderData(Dataset):
    def __init__(self, tsd: PandasTSDataSet, **params):
        self.tsd = tsd  # Store dataset reference

        # Access metadata from dataset (D1)
        self.metadata = tsd.get_metadata()

    def __getitem__(self, idx):
        sample = self.tsd[idx]  
        # other required additions to ``sample``
        return sample
```

In this way, we don't deviate from `setup` documentation, pass the metadata as well to other layers and we don't have to return the metadata in the `__getitem__`.

> Following up on this, I would actually put even more in single dicts, and use strings (like in pandas) for types:

Yeah! we can do that as well. Didn't strike me at that time. Thanks!

fkiraly:
@phoeenniixx, brilliant! Great thinking!

In restrospect - looking at  your pots, it looks like with layer D2 we were trying to reinvent the `LightningDataModule` interface!

Nice! I will need some time to digest, but I think we are getting close now.

FYI @agobbifbk, @felipeangelimvieira

phoeenniixx:
Thanks! @fkiraly 
> In restrospect - looking at your post, it looks like with layer D2 we were trying to reinvent the `LightningDataModule` interface!

Well I didn't fiind anyother way in `lightning` module that can handle this operation so we need to do something :)

phoeenniixx:
Thanks! @fkiraly 
> In restrospect - looking at your post, it looks like with layer D2 we were trying to reinvent the `LightningDataModule` interface!

Well I didn't fiind anyother way in `lightning` module that can handle this operation so we need to do something :)

agobbifbk:
Hello there! 
I'm trying to fit this structure in DSIPTS:

1.  Layer D1 PandasTSDataSet easy, it is almost a copy-paste from the original implementation. There are some problems in the ``__getitem__`` function since we can not pass strings (categorical variables) and timestamps! The first it is easy to solve, just use a LabelEncoder, the second one is more difficult, maybe again LabelEncoder? It may depend on the tipe of time we have (integers, dates, datetimes)...Moreover I thought useful to use the IDX of the getitem for retrieving only data of a certain **group**. This will hard affects the D2 layer implementation! But at least we are sure that one item of the D1 contains ONE time series. In the case we are reading from disk I suppose that the IDX will be another thing, like a chunk of data, that can have inside again the **group** that can be different among the chunks (for example group 1 can be or not in the chunk 1). I think that we can **precompute** the mapping chunk->groups (useful in the D2!)
 
3.  Layer D2: I'm finding some difficulties to think it working properly in the case of a non-PandasTSDataSet. For sure this snipped is not what we want
```python 

    def train_dataloader(self):
        return DataLoader(
            self.d1_dataset, #--> the dataloader then feeds the deep learning model, so we need to slice data before 
            batch_size=self.batch_size,
            **other_params
        )
```
One solution is to create a custom dataset and:
```python

    def train_dataloader(self):
        return DataLoader(
            MyDataset(self.d1_dataset), #--> custom dataloader
            batch_size=self.batch_size,
            **other_params
        )
```

where:
```python

class MyDataset(Dataset):

    def __init__(self,data,metadata)->torch.utils.data.Dataset:
       
        self.data = data  
        self.metadata = metadata
        self.metadata_dataset = data.metadata
    def __len__(self):
        
        return ?? ##hard to define now :-) 

    def __getitem__(self, idxs): ##this is the idx of the dataloader DEPENDS but it DEPENDS ON IDX ---> problem!
        sample = {}
        IDX = random.randint(0,len(ds)) ##THIS IS THE IDX OF THE TIMESERIES....RANDOM IS CORRECT???
        tmp = self.data.__getitem__(IDX)
        for k in self.data.keys:
            if k in self.metadata_dataset['past_variables']:
                sample[k] = tmp[k][idxs-self.metadata['past_steps']:idxs]
            if k in self.metadata_dataset['future_variables']:
                sample[k] = tmp[k][idxs:idxs+self.metadata['future_steps']]
        return sample
```
Before continuing I will study if and how it is managed inside PTF. 

Finally it is hard for me to understand where the scalers should act in the case of chunked data. But we will find a solution. If you have any comment please drop me a message!

agobbifbk:
> Hello there! I'm trying to fit this structure in DSIPTS:
> 
> 1. Layer D1 PandasTSDataSet easy, it is almost a copy-paste from the original implementation. There are some problems in the `__getitem__` function since we can not pass strings (categorical variables) and timestamps! The first it is easy to solve, just use a LabelEncoder, the second one is more difficult, maybe again LabelEncoder? It may depend on the tipe of time we have (integers, dates, datetimes)...Moreover I thought useful to use the IDX of the getitem for retrieving only data of a certain **group**. This will hard affects the D2 layer implementation! But at least we are sure that one item of the D1 contains ONE time series. In the case we are reading from disk I suppose that the IDX will be another thing, like a chunk of data, that can have inside again the **group** that can be different among the chunks (for example group 1 can be or not in the chunk 1). I think that we can **precompute** the mapping chunk->groups (useful in the D2!)
> 2. Layer D2: I'm finding some difficulties to think it working properly in the case of a non-PandasTSDataSet. For sure this snipped is not what we want
> 
> ```python
>     def train_dataloader(self):
>         return DataLoader(
>             self.d1_dataset, #--> the dataloader then feeds the deep learning model, so we need to slice data before 
>             batch_size=self.batch_size,
>             **other_params
>         )
> ```
> 
> One solution is to create a custom dataset and:
> 
> ```python
>     def train_dataloader(self):
>         return DataLoader(
>             MyDataset(self.d1_dataset), #--> custom dataloader
>             batch_size=self.batch_size,
>             **other_params
>         )
> ```
> 
> where:
> 
> ```python
> class MyDataset(Dataset):
> 
>     def __init__(self,data,metadata)->torch.utils.data.Dataset:
>        
>         self.data = data  
>         self.metadata = metadata
>         self.metadata_dataset = data.metadata
>     def __len__(self):
>         
>         return ?? ##hard to define now :-) 
> 
>     def __getitem__(self, idxs): ##this is the idx of the dataloader DEPENDS but it DEPENDS ON IDX ---> problem!
>         sample = {}
>         IDX = random.randint(0,len(ds)) ##THIS IS THE IDX OF THE TIMESERIES....RANDOM IS CORRECT???
>         tmp = self.data.__getitem__(IDX)
>         for k in self.data.keys:
>             if k in self.metadata_dataset['past_variables']:
>                 sample[k] = tmp[k][idxs-self.metadata['past_steps']:idxs]
>             if k in self.metadata_dataset['future_variables']:
>                 sample[k] = tmp[k][idxs:idxs+self.metadata['future_steps']]
>         return sample
> ```
> 
> Before continuing I will study if and how it is managed inside PTF.
> 
> Finally it is hard for me to understand where the scalers should act in the case of chunked data. But we will find a solution. If you have any comment please drop me a message!

Maybe I found a solution, I will post it as soon as possible!

agobbifbk:
Here https://github.com/agobbifbk/DSIPTS_PTF/blob/main/notebooks/TEST.ipynb you can find a working example with the proposed schema, maybe it can be optimized somehow. Not clear where and how to perform the scaling and how to align it with a different TSDastaset

fkiraly:
@agobbifbk, @phoeenniixx, interesting prototypes!

From discussion on Feb 17: we have noted that the two D1 layers have slightly different `__getitem__` return interfaces, we should try to align on something that satisfies requirements of both ptf and dsipts, and also some general requirements like "small number of tensors" and simplicity.

fkiraly:
> In the [sktime/pytorch-forecasting#1770](https://github.com/sktime/pytorch-forecasting/pull/1770), the "separation" into categorical, numerical etc are not done as explicitly as in https://github.com/agobbifbk/DSIPTS_PTF/blob/main/notebooks/TEST.ipynb, In the PR, what we do is save the columns in the metadata that which are categorical etc, and then do the splitting in the D2 layer based on that, in this way the number of tensors returned are less

Currently, I would indeed prefer that - because that way, the D2 layer (which may differ per model) can decide how it deals with categoricals, etc, and we minimize the number of tensors in D1.

What are the good reasons to keep this in D1? Genuine question.

agobbifbk:
> In the https://github.com/sktime/pytorch-forecasting/pull/1770, the "separation" into categorical, numerical etc are not done as explicitly as in https://github.com/agobbifbk/DSIPTS_PTF/blob/main/notebooks/TEST.ipynb, In the PR, what we do is save the columns in the metadata that which are categorical etc, and then do the splitting in the D2 layer based on that, in this way the number of tensors returned are less

I slightly disagree on that. If you filter out the unused columns in D1 you will have less data to pass to the D2....

> Currently, I would indeed prefer that - because that way, the D2 layer (which may differ per model) can decide how it deals with categoricals, etc, and we minimize the number of tensors in D1.

... But if we want to create a D2 per model  (I will say not per model but for **type** of model) maybe it can worth since in the D2 will get rid of unused columns. Maybe it will be convenient to separate  at least  numerical and categorical in the D1 so we can transform strings into integers directly at the beginning.

I will update my notebook according to this and try to return the less tensors as possible.

fkiraly:
> I slightly disagree on that. If you filter out the unused columns in D1 you will have less data to pass to the D2....

This sounds like a misunderstanding? I do not see any filtering happening in D1 design of @phoeenniixx. The categorical and numerical columns are simply kept in the same tensor (as floats). Please correct me if I am wrong, either of you.

> Maybe it will be convenient to separate at least numerical and categorical in the D1 so we can transform strings into integers directly at the beginning.

I agree with half of this - I do not think separating the tensors is necessary for a conversion of strings to integers. The columns that were categorical strings, after conversion to integers, can be stored in the same tensor as numerical variables, as long as we keep track in the metadata wihch columns are meant to be categorical.

I am not saying this is something we must do, I am just trying to make the point that there is no logical necessity to separate the tensors.

Beyond this, I agree with @thawn on the general point that, all other things being equall, less tensors in D1 `__getitem__`  are a better design.

agobbifbk:
Ok i got it! It make also sense to incorporate x and y and return a generic _data_ (or _x_) and the index of the target are stored in the metadata too.

agobbifbk:
> > Ok I got it! Can you please add @xandie985 to the thread? THX
> 
> which thread? He can simply reply here, if that is what you mean, this repository is public.

Ok sorry I did't find him using @ and autocompletition. We are working at the notebook you can find some progress here: http://localhost:8889/lab/tree/TEST_v2.ipynb Sandeep will take care of complete the missing part, after that we can compare this and the @phoeenniixx implementation!

fkiraly:
> > > Ok I got it! Can you please add @xandie985 to the thread? THX
> > 
> > 
> > which thread? He can simply reply here, if that is what you mean, this repository is public.
> 
> Ok sorry I did't find him using @ and autocompletition. We are working at the notebook you can find some progress here: http://localhost:8889/lab/tree/TEST_v2.ipynb Sandeep will take care of complete the missing part, after that we can compare this and the @phoeenniixx implementation!

two questions:

- I think you did correctly ping Sandeep on GitHub? That was xandie985 as you wrote above.
- the notebook link seems to your local, not to a remote

PranavBhatP:
Hi! @agobbifbk , you had mentioned in the last weekly-meeting about the requirement for an XArrayDataset to handle larger-than-memory dataset inputs. I'm not sure how appropriate this might be to the redesign and I'm open to comments on this, but I put some thought into it and came up with a rudimentary design with implementation for basic features of the Dataset, it currently inherits from PandasTSDataset design suggested in the TestV2.ipynb notebook. This would go with the D1 layer.

```python

class XArrayTSDataset(PandasTSDataset):
	def __init__(
		self,
		data: xr.Dataset, # todo: add zarr handling also
		time: optional[str] = None,
		target: Optional[Union[[str], List[str]]] = None,
		group: Optional[List[str]] = None,
		num: Optional[List[Union[str, List[str]]]] = None,
		cat: Optional[List[UNion[str, List[str]]]] = None,
		know: OPtional[List[Union[str, List[str]]]] = None,
		unknown: Optional[List[Union[str, List[str]]]] = None,
		static: Optiona[List[Union[str,List[str]]]] = None,
		label_encoder: Optional[dict] = None,
		chunk_size: Optional[Dict[str,int]] = None,
	):
		"""Initial thoughts and design on an XArrayTSDataset for larger-than-memory datasets, Inherits from PandasTSDataset as suggested in the notebook, but handles data using XArray for larger_than_memory dataset. Does well with pre-computing chunk indices.
		Additional Params
		-----------------
		chunk_size: Dict[str, int], optional
			Size of chunks for each dimension, e.g {'time':100, 'group' : 10}

		
	"""
		self.xr_data = data

		if chunk_size:
			self.xr_data = self.xr_data.chunk(chunk_size)

		first_chunk = self._get_first_chunk()
		
		super().__init__(
			data=first_chunk,
			time=time,
			target=target,
			group=group,
			num=num,
			cat=cat,
			known=known,
			unknown=unknown,
			static=static,
			label_encoders=label_encoders
		)
		self.chunk_size = chunk_size
		self._setup_chunks()

	def _get_first_chunk(self) -> pd.DataFrame:
		"""Get first chunk of data as pandas DataFrame."""

		if self.chunk_size > 0:
			subset = {}
			for dim, size in self.chunk_size.items()
				subset[dim] = slice(0,size)
			return self.xr_data.isel(**subset).to_dataframe()
		
	def _setup_chunks(self):
		"""Setup chunk information for iteration on init.."""
		if not self.chunk_size:
			self.n_chunks = 1
			return

		self.chunks_per_dim = {}
		for dim in self.chunk_size:
			total_size = self.xr_data.sizes[dim]
			chunk_size = self.chunk_size[dim]
			self.chunks_per_dim[dim] = int(np.ceil(total_size /  chunk_size))

		self.n_chunks = np.prod(list(self.chunks_per_dim.values())) ##Gets the number of chunks based on the dim values.

		self.chunk_indices = list(itertools.product(
			*[range(n) for n in self.chunks_per_dim.values()]
		))

	def _get_chunk(self, chunk_idx: int) -> xr.Dataset:
		"""Getter function for chunks, gets the chunk_coords for chunk based on the chunk_idx"""

		chunks_coord = self.chunk_indices[chunk_idx]

		chunk_slice = {}

		for dim_idx, dim in enumerate(self.chunk_size.keys()):
			start = chunk_coords[dim_idx] * self.chunk_size[dim]
			end = min(start + self.chunk_size[dim], self.xr_data.sizes[dim]) 
			chunk_slices[dim] = slice(start, end)

		return self.xr_data.isel(**subset)


	def __getitem__(self, index: int) --> Dict[str, torch.Tensor]:
		"""Get time series data for a given index."""

		items_per_chunk = np.prod([self.chunk_size[dim] for dim in self.chunk_size])
		# This is a simple calculation for getting the 
		# chunk number of the current chunk.
		chunk_idx = index // items_per_chunk

		# offset
		local_idx = index % items_per_chunk

		chunk_data = self._get_chunk(chunk_idx)

		chunk_df = chunk_data.to_dataframe()

		return super().__getitem__(local_idx)

	def __len__(self) --> int:
		if not self.chunk_size:
			return super().__len__()
		total_items = np.prod([
			self.xr_data.sizes[dim] for dim in self.chunk_size
		])
		return total_items


	def _process_chunk(self, chunk: xr.Dataset) -> Dict[str, torch.Tensor]:
		"""Function to process chunks into Tensors"""

	def to_pandas(self) -> pd.DataFrame:
		"""Dumps the xarr data into memory, just including some functions to show the possible usecases for this class."""
```

agobbifbk:
Nice! I think a lot of stuff can not be inherited from PandasTSDataset (for example label encoder) but it is a starting point. We can implement it as separate class and then see if there are common part (in that case we can have a generic Dataset and then the pandas and xarray will derive from that).
Can you also add an example where you have a pandas dataframe and convert it to a xarray chunked by time and groups? You can start from TEST_v2 dataset and try to work around it! THX

xandie985:
### Code Update: MultiFileTSDataset for Efficient Multi-File Time Series Handling

In this code we introduce the class  `MultiFileTSDataset`, designed to process time series data from multiple CSV files, optimized for large datasets.

**Key Features:**

* **Multi-File & Chunk Processing:** Handles multiple CSV files via `file_paths`, reading data in `chunk_size` chunks to manage memory. **Crucial point to discuss around:** Input CSVs must be pre-sorted by `time_col` to maintain temporal order.
* **Global Encoding & Grouping:** Uses a shared `OrdinalEncoder` for categorical and group columns (`cat_cols`, `group_cols`). Supports grouping, storing group-specific data in `group_info`.
* **Time-Based Weighting:** Implements exponential decay weighting based on time differences from the most recent sample via `_calculate_weight`. (we can use other time based weighing mechanisms as well. )
* **Efficient Data Retrieval:** `__getitem__` locates and loads data using `file_info` and `_load_chunk`, returning time, target, encoded features, and weights as tensors.
* **Chunk Loading:** `_load_chunk` loads data chunks and filters by group if needed.

**Considerations:**

* **Data Sorting:** Input files MUST be sorted by the time column. 
* **Memory Tuning:** Adjust `chunk_size` for optimal performance.
* **Timestamp Conversion:** time column is converted to timestamp for consistency.
* **Weight decay rate:** the decay rate can be tuned.

### Code:

```python
class MultiFileTSDataset(Dataset):
    """PyTorch Dataset for handling multiple files with different groups.
    
    This class processes multiple files containing time series data with different
    groups, maintains a global label encoder, and handles data in chunks.
    """
    
    def __init__(
        self,
        file_paths: List[str],
        time_col: str,
        group_cols: List[str],
        target_cols: List[str],
        cat_cols: List[str],
        chunk_size: int = 1000 
    ):
        """
        Parameters
        ----------
        file_paths : List[str]
            List of paths to CSV files
        time_col : str
            Name of the time column
        group_cols : List[str]
            Columns to group by
        target_cols : List[str]
            Target columns to predict
        cat_cols : List[str]
            Categorical columns to encode
        chunk_size : int
            Number of rows to process at a time
        """
        self.file_paths = file_paths
        self.time_col = time_col
        self.group_cols = group_cols
        self.target_cols = target_cols
        self.cat_cols = cat_cols
        self.chunk_size = chunk_size #on what basis chunking should be done while reading the data?
        # temporal information might get lost during chunking if user data is not sorted by default.
        
        #  label encoders for categorical and group columns
        self.label_encoders = {col: OrdinalEncoder(handle_unknown='use_encoded_value', unknown_value=-1) 
                              for col in cat_cols + group_cols}
        
        # lists to store file and group information
        self.file_info = []
        self.group_info = {}
        
        # Process files and build index
        self._process_files()
        
    def _process_files(self):
        """Process each file to extract group information and update encoders."""
        self.total_length = 0  #init total length of the dataset
        
        for file_idx, file_path in enumerate(self.file_paths):
            # Read each file in chunks
            for chunk in pd.read_csv(file_path, chunksize=self.chunk_size):
                # updating label encoders with new categories from the chunk
                self._update_encoders(chunk)
                
                # If group columns are specified
                if self.group_cols:
                    # unique groups in the chunk
                    groups = chunk[self.group_cols].drop_duplicates()
                    
                    for _, group_row in groups.iterrows():
                        # Create a key from the group columns' values
                        group_key = tuple(group_row[self.group_cols].values)
                        
                        # If this group is not seen before, initialize its entry in group_info
                        if group_key not in self.group_info:
                            self.group_info[group_key] = []
                        # Add the current file index and file info index to the group's entry
                        self.group_info[group_key].append((file_idx, len(self.file_info)))
                        
                        # Filter data for the current group
                        group_mask = (chunk[self.group_cols] == group_row).all(axis=1)
                        group_data = chunk[group_mask]
                        
                        # Store information about the group in file_info
                        self.file_info.append({
                            'file_idx': file_idx,
                            'group_key': group_key,
                            'min_time': group_data[self.time_col].min(),
                            'max_time': group_data[self.time_col].max(),
                            'length': len(group_data)
                        })
                        
                        self.total_length += len(group_data)  # updating total length
                else:
                    # if no group columns -> store file information
                    self.file_info.append({
                        'file_idx': file_idx,
                        'group_key': None,
                        'min_time': chunk[self.time_col].min(),
                        'max_time': chunk[self.time_col].max(),
                        'length': len(chunk)
                    })
                    
                    self.total_length += len(chunk)  # Update total length
    
    def _update_encoders(self, data):
        """Update label encoders with new categories from the data."""
        for col in self.cat_cols + self.group_cols:  # for each categorical and group column
            if col in data.columns:  # If the column exists in the data
                unique_values = data[col].dropna().unique().reshape(-1, 1)  # Get unique values
                if len(unique_values) > 0:  # If there are unique values
                    self.label_encoders[col].fit(unique_values)  # update the encoder for this column
    
    def __len__(self):
        """Return the total number of items in the dataset."""
        return self.total_length
    
    def _calculate_weight(self, timestamp):
        """Calculate weight based on time difference from most recent sample."""
        # Convert timestamp to datetime if needed
        if isinstance(timestamp, (int, float)):
            timestamp = pd.to_datetime(timestamp, unit='s')
        elif isinstance(timestamp, str):
            timestamp = pd.to_datetime(timestamp)
        
        # Get most recent time in dataset
        most_recent = max(info['max_time'] for info in self.file_info)
        
        # Calculate time difference in days
        time_diff = (most_recent - timestamp).days
        
        # Apply exponential decay for weights; can adjust the decay rate (0.1) as needed
        weight = np.exp(-0.1 * time_diff)
        
        return weight

    def __getitem__(self, idx):
        """Get a single item from the dataset."""
        # Find the file and group that the given index belongs to
        cumulative_length = 0
        for info in self.file_info:
            if idx < cumulative_length + info['length']:
                file_idx = info['file_idx']
                group_key = info['group_key']
                local_idx = idx - cumulative_length
                break
            cumulative_length += info['length']
        
        # Load the relevant chunk from the file
        chunk = self._load_chunk(file_idx, group_key)
        
        # Get the specific row from the chunk
        row = chunk.iloc[local_idx]
        
        # Encode categorical columns
        encoded_data = {}
        for col in self.cat_cols + self.group_cols:
            if col in row.index:
                encoded_data[col] = self.label_encoders[col].transform([[row[col]]])[0][0]
        
        # Convert time to numeric timestamp
        time_value = pd.to_datetime(row[self.time_col]).timestamp()
        
        # Extract target values and ensure they are numeric
        target_values = []
        for col in self.target_cols:
            target_values.append(float(row[col]))
        
        # Convert data to PyTorch tensors 
        result = {
            "t": pd.to_datetime(row[self.time_col]).timestamp(), # NOT TENSOR 
            "y": torch.tensor(target_values, dtype=torch.float32),
            "x": torch.tensor(list(encoded_data.values()), dtype=torch.float32),
            "weights": torch.tensor(self._calculate_weight(row[self.time_col]), dtype=torch.float32)
}
     
        return result
    
    def _load_chunk(self, file_idx, group_key):
        """Load a chunk of data from a specific file and group."""
        # Load the next chunk from the specified file
        chunk = pd.read_csv(self.file_paths[file_idx], chunksize=self.chunk_size).__next__()
        
        # Convert the time column to datetime format
        chunk[self.time_col] = pd.to_datetime(chunk[self.time_col])
        
        # If a group key is specified, filter the chunk by that group
        if group_key is not None:
            group_mask = True
            for i, col in enumerate(self.group_cols):
                group_mask = group_mask & (chunk[col] == group_key[i])
            return chunk[group_mask]
        return chunk

**Comment by agobbifbk** [comment]
Thx @PranavBhatP. I have some comments:

1. I think that a general way for passing optimizer, scheduler and so one is to use the `eval` function and pass the module string `torch.optim.Adam` otherwise we are limited only to use `ADAM` and `SGD` , same for `activation` I see this:
```python
self.activation = F.relu if activation == "relu" else F.gelu
```
I prefer to let the user pass the activation function or define a default (ReLU). This a more general concept, you will find a lot of this example while exploring other library, I think it worth to raise the question during a technical session
2.       
```python
encoder_target = x.get("target", torch.zeros(batch_size, self.prediction_length, 0, device=self.device))
``` 
We need to investigating this. Is `target` the `y`? The variable we want to predict? Why is it used during the training phase? This is something strange IMHO, maybe what you want is the `y` in the past. So we need to add this information in D2 layer. We have 2 ways:
- pass the `past_y` separately (need to change **D2** layer)
- pass the autoregressive component in `x` and the array of index in which we have the autoregressive part (need to change **D2** layer)
Since this autoregressive part is used in several models we need to add this to the D2 layer. I see you are defining `target_pos` and not using it, maybe it is something you planned to add :-) 

3. Do you have some information about `encoder_time_idx`:
```python
 encoder_time_idx = x.get("encoder_time_idx", None) 
```
maybe it is worth to add it to the D2?
4. `encoder_cat` and the `decoder` part is not used, it seems a waste of information...Some ideas on that:

- add an `Embedding layer` to process `cat` variables (more information [here](https://pytorch.org/docs/stable/generated/torch.nn.Embedding.html)). This implies to compute the input size in the **D1** layer (the cardinality of each categorical variable)
- the decoder part should be added somewhere near `dec_out = self.head(enc_out)`. Here all the information of the past has been processed. We can add some linear `after` the head layer of `before`. This is a matter of dimension!

5. Not clear the output dimension. Does it work for multioutput time series? If quantile loss is used I expect to have somewhere a `*3` on the output channels (usually Quantile uses 0.05, 0.5, 0.95 values this is why I say `*3`)

Concluding: once the D2 layer is completed, importing models from THUML seems a matter of notation and standardization. With a little of effort we can:
-  add missing information (aka: decoder part), In this case maybe is better to  use a flag `use_original` to use the original implementation?
- abstract optimizer, schedulers, activation using `eval` and the method. I think it will be a more elegant solution, what do you think? 

Thank you for the effort!! Have a nice rest of the day

PR #43, opened 2025-07-01
Advanced Benchmarking

jgyasu:
Proposes dataset collections for simplifying benchmarking experiment reproducibility.

### DIFF
--- a/steps/25_dataset_collection/step.md
+++ b/steps/25_dataset_collection/step.md
@@ -1,4 +1,4 @@
-# Collection of Datasets
+# Advanced Benchmarking
 
 Contributors: ["jgyasu"]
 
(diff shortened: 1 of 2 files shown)


steps/25_dataset_collection/step.md
```diff
@@ -0,0 +1,158 @@
...
+            dataset = dataset_loader()
+            self._additional_datasets.append(dataset)
+        except Exception as e:
+            print(f"Warning: Failed to add additional dataset: {e}")
+        return self
+
+    def __len__(self):
+        return len(self._collection.get_datasets()) + len(self._additional_datasets)
+
+    def get_collection_info(self):
```
felipeangelimvieira, line 153: Could this be part of a tag system for the dataset collection?

steps/25_dataset_collection/step.md
```diff
@@ -0,0 +1,158 @@
...
+    """Collection of TSC Bake-off datasets."""
+    
+    def __init__(self, subset=None):
+        """
+        Initialize TSC Bake-off collection.
+        """
+        self.subset = subset
+        super().__init__()
+    
+    def get_collection_name(self):
```
fkiraly, line 100: `get_collection_name` feels like it should be a tag, e.g., `"info:name" : "TSC Bake-off 2017`

steps/25_dataset_collection/step.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+```python
+
+class DatasetCollection(BaseObject):
+    """
+    A collection of datasets following the strategy pattern.
+    """
+
+    _collections = {
```
fkiraly, line 124: this design is not very extensible!

Why: for adding a collection, you need to:

1. add a new object
2. add it to this register

Generally, designs requiring a registry are less extensible than designs using the strategy pattern only.

fkiraly (changes requested):
Great proposal, I like the high-level ideas and design!

Some issues and comments I have:

* the `DatasetCollection` is a kind of register, that also functions as a instance factory (via `add`). I feel this is a violation of the single responsibility principle and the strategy pattern as well. So I would go for sth different.
* The use case of adding one more dataset is a nice one, I did not have it on the radar that clearly.
* the `add` method belongs to the builder pattern, and violates the dataclass pattern.  There is a way to have both: instead of modifying `self`, the `add` method returns a newly constructed `CustomCollection` (or similar), which is a dataclass.

"What if" ideas:

* how about collections that are not just for datasets, but anything? It could be estimators! Example: the TSC bake-off had a collection of datasets, and a collection of estimators as well.
* instead of using a registry in `DatasetCollection`, should we use `craft` from `registry`? This is a lookup entry point which gets class by name.

jgyasu:
> how about collections that are not just for datasets, but anything? It could be estimators! Example: the TSC bake-off had a collection of datasets, and a collection of estimators as well.

I will adress the other comments, but this is a nice idea. I thought about it too. I think we can just hace a base class for collections, say, `BaseCollection` and then extend it to different types of collections, `EstimatorBaseCollection`, `DatasetBaseCollection`.

If this sounds good then I will modify this proposal? @fkiraly

fkiraly:
Sounds good, although I would only introduce subclasses if something implies they are needed, design-wise.

PR #47, opened 2026-02-11
[STEP] Design Document for rework of ForecastingHorizon

RecreationalMath:
[WIP] A proposal for the rework of `ForecastingHorizon` and a potential decoupling from `pandas.Index`
[sktime/Issue#7617](https://github.com/sktime/sktime/issues/7617)

Note: The Jupyter file `sktime_fh.ipynb` shall be deleted before merging the STEP PR as it contains code pertaining to observation of `ForecastingHorizon`'s behavior before the rework.


RecreationalMath:
FYI @fkiraly @SimonBlanke 
Please share your inputs on some of the questions raised in the STEP document.
I have shared my concerns on certain choices/trade-offs but haven't made a recommendation yet, as a lot hinges on 
1. `sktime`'s need for having converters to and from pandas objects for interoperability between various modules and not just within `ForecastingHorizon`. (if we remove the converters from `ForecastingHorizon`, we still would still need them for interoperability, right?)
2. Backward compatibility of `ForecastingHorizon` API v/s breaking away from it.

RecreationalMath:
> 2. Backward compatibility of `ForecastingHorizon` API v/s breaking away from it.

FK: I think we can go either way - have a refactor target and then deprecate rarely used methods. I think most users will not use it directly, even though it is public. For 1.0, we can introduce breaking changes if it feels beneficial to the user experience and design; but all other things equal we should avoid breaking changes.

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+
+1. Interface Compatibility & Backward Compatibility
+
+    - Q1.1. Whether the new class should be a drop-in replacement that maintains 100% API compatibility with current implementation, because end users already use it and expect it to behave in a certain way? And changing it might need following deprecation patterns and introduce complexities. Or a refactor of the interface itself can also be considered? 
+    - Consideration1: Assuming some (many?) users would be using the exposed methods of `ForecastingHorizon` in their workflows and preserving them from breaking by having the same methods still exist after the rework, is of importance. 
+    - Consideration2: Changing the interface would require maintaining both classes for a while and a complicated deprecation period.
+    - Conflict: 
+        - Keeping same interface means `ForecastingHorizon` accepts the pandas objects as inputs. Which puts the responsibility to manage the `freq` mnemonics changes by `pandas` on `sktime`. Same problem as before.
```
fkiraly, line 94: ... yes, but optimally the problem could be isolated to an adapter/coercer class or funtion, or, more generally, isolated "away" architecturally from the new `ForecastingHorizon`.

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+        - If we want to ensure absolutely no breakage with regards to `freq` mnemonics change, the onus of converting to a `ForecastingHorizon` compatible input type needs to be transfered to the end-user. `sktime` can obviously help by provide a utility for conversion. But if we do two things will happen.
```
fkiraly, line 95: maybe - alternatively, we can attempt to minimize the onus by localizing the `freq` logic - and its treatment across different versions - into a pure conversion step that is highly focalized.

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+    - Conflict: 
+        - Keeping same interface means `ForecastingHorizon` accepts the pandas objects as inputs. Which puts the responsibility to manage the `freq` mnemonics changes by `pandas` on `sktime`. Same problem as before.
+        - If we want to ensure absolutely no breakage with regards to `freq` mnemonics change, the onus of converting to a `ForecastingHorizon` compatible input type needs to be transfered to the end-user. `sktime` can obviously help by provide a utility for conversion. But if we do two things will happen. 
+            1. current working workflows of end-users will break. We can use deprecation and handle it gracefully.
+            2. yt will add one extra step before/after using the sktime forecasters -> to convert the forecasting horizon from pandas to fh_compatible input types. Seems tedious, need inputs and thoughts.
+    
+
+2. Internal Data Representation
+
+    - Q2.1 Should the class still support all input types (int, list, np.ndarray, pd.Index, timedelta, date offsets)?
```
fkiraly, line 108: I would say, yes - but via conversion to a crisp internal representation, rather than internally.

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+    - Consideration: Same concern as above. If we support the pandas objects as input types, there would always be a chance of breakages whenever pandas makes changes as it currently happens. (When our convertor tries to convert an unseen freq type to map it to the internal representation of the custom `fh`)
+    - Q2.2: Instead of wrapping pd.Index, what should be the primary internal representation?
+    - Option A: numpy arrays for numeric values + custom metadata objects (least coupling with pandas)
+    - Option B: WILD IDEA!! what if instead of de-coupling we embrace the pandas fully and don't write converters but use pandas objects directly? Need investigation on what changes would that wrrant.
```
fkiraly, line 116: strongly disagree. Bad idea imo, because:

* still would break on the interfaces
* how would you represent relative indices?

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+            1. current working workflows of end-users will break. We can use deprecation and handle it gracefully.
+            2. yt will add one extra step before/after using the sktime forecasters -> to convert the forecasting horizon from pandas to fh_compatible input types. Seems tedious, need inputs and thoughts.
+    
+
+2. Internal Data Representation
+
+    - Q2.1 Should the class still support all input types (int, list, np.ndarray, pd.Index, timedelta, date offsets)?
+    - Consideration: Same concern as above. If we support the pandas objects as input types, there would always be a chance of breakages whenever pandas makes changes as it currently happens. (When our convertor tries to convert an unseen freq type to map it to the internal representation of the custom `fh`)
+    - Q2.2: Instead of wrapping pd.Index, what should be the primary internal representation?
+    - Option A: numpy arrays for numeric values + custom metadata objects (least coupling with pandas)
```
fkiraly, line 115: yes, I would say option A is the way to go.

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+    - Option B: WILD IDEA!! what if instead of de-coupling we embrace the pandas fully and don't write converters but use pandas objects directly? Need investigation on what changes would that wrrant.
+
+    
+3. Pandas Interoperability (between different parts of sktime)
+
+    - Q3.1 sktime forecasters return pandas Series/DataFrames with pandas Index objects? Does that mean the new class still needs to provide `to_pandas()` and `to_numpy()` methods for integration?
```
fkiraly, line 123: that is a good question. This functionality could in theory also be provided by `BaseForecaster`, e.g., via an extension to `_get_expected_pred_idx`

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+    - Accept pandas Index objects as inputs but convert them immediately to internal representation? This would result in same concerns.
+
+4. Type Handling & Frequency Management
+
+    - Q4.1 For relative vs. absolute distinction, what are the tradeoffs between keeping the current type-based approach (inferring from value types) v/s using explicit metadata/flags instead?
```
fkiraly, line 128: I think infer from type allows to avoid to request from the user to always provide.

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+    - Q4.2 For frequency management (the complex freq property with setter) since a lot of breakages happen here, is there a way to make it optional instead of required? Will it break anything?
+
+5. MultiIndex support
+    - This is a new feature request. Can be done but the approach will depend on answers to other questions, i.e. how much away or close the picked end-state is to pandas.
```
fkiraly, line 135: yes, this is an interesting question. I think we should plan this with multiindex support in mind - namely, the ability to select different horizons for different time series in a hierarchical series or collection.

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+
+6. Performance & Caching
+
+I'm a bit rusty on my threading/caching/memory-leakages concepts and have used AI coding assistant's help in deliberating this part. Would appreciate inputs here. 
+Please feel free to validate/invalidate below points.
+
+- Hashability concerns: 
+    - aren't pandas Index objects hashable?
```
fkiraly, line 143: no. Go and try.

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+    - This is a new feature request. Can be done but the approach will depend on answers to other questions, i.e. how much away or close the picked end-state is to pandas.
+
+6. Performance & Caching
+
+I'm a bit rusty on my threading/caching/memory-leakages concepts and have used AI coding assistant's help in deliberating this part. Would appreciate inputs here. 
+Please feel free to validate/invalidate below points.
+
+- Hashability concerns: 
+    - aren't pandas Index objects hashable? 
+    - current code uses _HashIndex wrapper, this is also source of coupling with pandas.
```
fkiraly, line 144: exists because index is not hashable.

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,138 @@
...
+- Hashability concerns: 
+    - aren't pandas Index objects hashable? 
+    - current code uses _HashIndex wrapper, this is also source of coupling with pandas. 
+- Cache and thread safety concerns: 
+    - if many instances are created - is this implementation memory safe? 
+    - does lru_cache exhibit thread-safety?
+
+7. Maintenance load
+
+it seems edge-cases (and corresponding workarounds) will always be needed as long as there exists a code internal to `ForecastingHorizon` which converts to or from pandas objects. Unless that piece is specifically moved out and exposed to end-user and forcing them to make it part of their forecasting pipeline. Even after doing so, it will still break, but the fault localisation for the end-user might be easier and some heavy users might choose to write their own wrappers on top of `sktime` provided convertor utility. Giving end-users a way to fix their broken pipelines and also maybe sharing the blame for breakages with pandas.
```
fkiraly, line 153: I would go with the following approach: Introduce an additional conversion layer that handles pandas indices.  localize conversion out of the "core" functionality, and collect all `pandas` related logic in the conversion layer.

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+7. Maintenance load
+
+it seems edge-cases (and corresponding workarounds) will always be needed as long as there exists a code internal to `ForecastingHorizon` which converts to or from pandas objects. Unless that piece is specifically moved out and exposed to end-user and forcing them to make it part of their forecasting pipeline. Even after doing so, it will still break, but the fault localisation for the end-user might be easier and some heavy users might choose to write their own wrappers on top of `sktime` provided convertor utility. Giving end-users a way to fix their broken pipelines and also maybe sharing the blame for breakages with pandas. 
+
+## Description of proposed solution
+
+`ForecastingHorizon` to be split into three classes:
+1. `ForecastingHorizonV2` - the new class with the same interface as `ForecastingHorizon` but with an internal representation decoupled from pandas. It will be a drop-in replacement for `ForecastingHorizon` and will be the main class going forward.
```
fkiraly, line 158: I would give it the same name. If it is a drop-in replacement, we can drop in immediately starting release 1.0

steps/28_forecasting_horizon/step.md
```diff
@@ -0,0 +1,205 @@
...
+2. `FHValues` - a new base class that handles the functionality currently outsourced to pandas. It will be used internally by `ForecastingHorizonV2` to manage the values and frequencies.
+3. `PandasFHConverter` - a separate utility class that handles the conversion between pandas objects and the internal representation used by `ForecastingHorizonV2`. This class will be responsible for managing the coupling with pandas and will be the only part of the code that interacts directly with pandas objects.
```
fkiraly, line 160: I think this split makes sense. I wonder if we could do with two classes only, i.e., merge 1 and 2, but I think it is sensible to start with 3 and then see how it pans out.

RecreationalMath: I'll share my design considerations below but please feel free to share if there is a specific argument/thought behind merging, which I am missing.

FHValues is a clean, domain-agnostic container: "sorted, deduplicated int64 array +some metadata" It has no knowledge of forecasting concepts like "relative vs absolute" or "cutoff." This makes it reusable:

- it can be utilized for cutoff representation, else cutoff would need it's own implementation
- MultiIndex support: one idea is it could be done via a dict of `FHValues`, which is trivial if `FHValues` is its own class.
- general future reusability: in future if we need a base sequence type FHValues can be reused
- any future refactor where a base sequence type needs to be abstracted out to decouple the internal representation contracts from a class representing a specific real world time series concept.

While reusability is the main reasoning behind this design but there are some secondary advantages as well:
- single responsibility per class: If 1 and 2 are merged, `ForecastingHorizon` becomes responsible for both low-level array storage and high-level forecasting semantics. Not a great design in my opinion.
- testability and maintenaince: with this design, testing `FHValues` is independent of pandas or forecasting horizon, without any instances of them required. I believe this lessens the future maintenaince load.

steps/28_forecasting_horizon/step.md
RecreationalMath: > I would give it the same name. If it is a drop-in replacement

Makes sense, shall do if it remains so till the refactor design and implementation is complete. (we might run into an issue that makes us consider not keeping it so, though I don't foresee any which can't be handled)

steps/28_forecasting_horizon/step.md
RecreationalMath: > FHValues is a clean, domain-agnostic container: "sorted, deduplicated int64 array +some metadata" It has no knowledge of forecasting concepts like "relative vs absolute" or "cutoff." This makes it reusable:

Maybe the name (FHValues) can be changed to remove the notion that it is purely related to FH(?) 
Any suggestions?

RecreationalMath: Rename `FHValues` to `BaseTimeSequence` (placeholder name).

`_fh_utils.py` handles the usage of `BaseTimeSequence` for forecasting horizon needs.
`_someOtherSemantic_utils.py` handles the of `BaseTimeSequence` for someOtherSemantic needs.

PR #50, opened 2026-05-10
[STEP] Pretrain `set_params`, `reset` and `clone`

felipeangelimvieira:
This pull request adds a detailed design document describing the proposed API and requirements for handling pretrained parameters in global forecasters and pretraining-compatible models. The document outlines current problems, requirements, and proposed solutions for managing pretrained parameters during operations like `reset`, `clone`, and `set_params`, and suggests changes to both the API and underlying implementation to address edge cases and ensure robust, user-friendly behavior.

Please read the file in this PR for full details.

Related issues:
https://github.com/sktime/skbase/issues/554
https://github.com/sktime/sktime/issues/10155
https://github.com/sktime/sktime/issues/10151

### DIFF
--- a/steps/27_pretrain_parameters/step.md
+++ b/steps/27_pretrain_parameters/step.md
@@ -7,8 +7,8 @@ and pretraining-compatible models with respect to pretrained parameters.
 
 
 ## Related issues
-* https://github.com/sktime/sktime/issues/10151 
-* https://github.com/sktime/skbase/issues/554   
+* https://github.com/sktime/sktime/issues/10151
+* https://github.com/sktime/skbase/issues/554
 
 
 ## Problem
@@ -18,7 +18,7 @@ and pretraining-compatible models with respect to pretrained parameters.
 Currently, `set_params` resets all model parameters, including parameters learned
 during pretraining. This should be avoided because, when pretrained parameters
 are used in pipelines or other compositions, clone operations and `set_params`
-should keep those parameters intact, for example model weights. Therefore,
+should keep those parameters intact, for example, model weights. Therefore,
 similarly to the changes in the clone API, where pretrained parameters were
 made to remain unchanged, `set_params` should preserve pretrained parameters as
 well.
@@ -45,26 +45,49 @@ default behavior of preserving non-task-specific parameters.
 
 ### Problem 2: incompatibility of certain hyperparameters with existing pretrained parameters
 
-Consider the case where an user has a pretrained mode with certain hyperparameters (say `hidden_dim`), they set it to new values but, as the model is a global model, the pretrained parameters keep the old `hidden_dim` weights. This can cause errors when the user tries to use the model with the new hyperparameters, as the pretrained parameters are not compatible with the new hyperparameters.
+Consider the case where a user has a pretrained model with certain
+hyperparameters, such as `hidden_dim`, and sets them to new values. Because the
+model is a global model, the pretrained parameters keep the old `hidden_dim`
+weights. This can cause errors when the user tries to use the model with the new
+hyperparameters, because the pretrained parameters are not compatible with the
+new hyperparameters.
 
-We can make the user responsible for properly setting the correct parameters, but this could introduce errors that are hard to debug. Therefore, we should also make the user aware when they are trying to set incompatible hyperparameters with the existing pretrained parameters, and decide if we want to:
+We can make the user responsible for setting the correct parameters properly,
+but this could introduce errors that are hard to debug. Therefore, we should
+also make the user aware when they are trying to set hyperparameters that are
+incompatible with the existing pretrained parameters, and decide whether we want
+to:
+
 1.  raise an error in this case
-2.  just a warning and force the reset of pretrained parameters when incompatible hyperparameters are set. 
+2.  raise a warning and force the reset of pretrained parameters when
+    incompatible hyperparameters are set.
 
-The option `2.` can silently change the behavior of the model without the user being fully aware of it (if they are not checking the logs). Therefore, we consider option 1. that is more explicit and forces the user to be aware of the exact behavior of the code, while avoiding silent buggy behavior. 
+Option 2 can silently change the behavior of the model without the user being
+fully aware of it, if they are not checking the logs. Therefore, we consider
+option 1 to be more explicit, because it makes the user aware of the exact
+behavior of the code while avoiding silent buggy behavior.
 
-This would required models to check if a certain change in `__init__` arguments is incompatible with the existing pretrained parameters, and raise an error in this case. This might not be linear since, for certain parameters, the incompatibility of the change might depend on the value of other parameters and current state.
+This would require models to check whether a certain change in `__init__`
+arguments is incompatible with the existing pretrained parameters, and to raise
+an error in that case. This might not be linear because, for certain parameters,
+the incompatibility of the change might depend on the value of other parameters
+and the current state.
 
 ### Problem 3: shared memory between cloned objects
 
-When pretrained parameters are not reset during cloning, there is a risk of shared memory between the original and cloned objects. This can lead to unintended side effects when one of the objects is modified, as it can affect the other object as well. Therefore, we need to ensure that when pretrained parameters are not reset during cloning, they are also properly copied to avoid shared memory issues.
+When pretrained parameters are not reset during cloning, there is a risk of
+shared memory between the original and cloned objects. This can lead to
+unintended side effects when one of the objects is modified, as it can affect
+the other object as well. Therefore, we need to ensure that when pretrained
+parameters are not reset during cloning, they are also properly copied to avoid
+shared memory issues.
 
 ## Requirements
 
 * by default, `reset` should have the same behavior as `clone`
-* non-task-specific parameters should be cloned (or deep-copied if `.clone()` is not supported) and kept their values unchanged by default in `reset`,  `clone`, and `set_params` operations
+* non-task-specific parameters should be cloned (or deep-copied if `.clone()` is not supported) and keep their values unchanged by default in `reset`, `clone`, and `set_params` operations
 * users should be able to reset non-task-specific parameters when needed
-* users should be made aware when they are trying to set incompatible hyperparameters with the existing pretrained parameters, and an error should be raised in this case.
+* users should be made aware when they are trying to set hyperparameters that are incompatible with the existing pretrained parameters, and an error should be raised in this case.
   
 ## Proposed solution
 
@@ -80,34 +103,36 @@ When pretrained parameters are not reset during cloning, there is a risk of shar
 3. `reset` and `clone` should both inspect that class attribute, reset the
    listed attributes if `force_reset=True`, and keep them cloned/deep-copied if
    `force_reset=False`. Every other attribute should be reset as usual.
-4. `set_params` should call a function `_check_set_params_compatibility(self, **params) -> None` that checks whether the new hyperparameters are compatible with the existing pretrained parameters, and raises an error if they are not. This function should be called before setting the new parameters. See the following skbase changes for more details on how to implement.
+4. `set_params` should call a function `_check_set_params_compatibility(self, **params) -> None` that checks whether the new hyperparameters are compatible with the existing pretrained parameters, and raises an error if they are not. This function should be called before setting the new parameters. See the following skbase changes for more details on how to implement this.
 
-### Skbase changes 
+### Skbase changes
 
 For handling incompatible hyperparameters with existing pretrained parameters:
-* Option 1: Add `set_params` plugin as in `clone`
-* Option 2: explicitely call a function `_check_set_params_compatibility(self, **params) -> None` in the base `set_params` implementation.
+* Option 1: Add a `set_params` plugin as in `clone`
+* Option 2: explicitly call a function `_check_set_params_compatibility(self, **params) -> None` in the base `set_params` implementation.
 
 For handling non-resettable attributes:
 
-* On skbase side, we could add for objects a `_NON_RESETTABLE_ATTRIBUTES_ATTR_NAME` class attribute that specifies the name of the class attribute that lists non-resettable attributes. 
-* Default is `None` and maintains the current behavior of not having non-resettable attributes. 
+* On the skbase side, we could add a `_NON_RESETTABLE_ATTRIBUTES_ATTR_NAME` class attribute for objects. This attribute specifies the name of the class attribute that lists non-resettable attributes.
+* The default is `None`, which maintains the current behavior of not having non-resettable attributes.
 * If set to a string, it should be the name of a class attribute that lists non-resettable attributes.
-* During reset and clone operations, it should check if the attribute exists and throw an error if it is not set. It should use `force_reset` argument to then decide whether to reset the attributes listed in that class attribute or not. Passing `force_reset` to objects that do not have non-resettable attributes should not cause any issues and behave as usual.
+* During reset and clone operations, it should check whether the attribute exists and raise an error if it is not set. It should then use the `force_reset` argument to decide whether to reset the attributes listed in that class attribute. Passing `force_reset` to objects that do not have non-resettable attributes should not cause any issues and should behave as usual.
 
-This would allow us to have a more general API for estimators that have attributes that are not task-specific
+This would allow us to have a more general API for estimators that have attributes that are not task-specific.
 
 
 
 ## Alternative solutions
 
 ### Use only the function argument `_reset` or `reset_pretrained`
+
 This would require fewer API changes, but it would not solve the problems in
 `reset` and would make it hard to control the compatibility of global models in
 pipelines and other compositions (see issue
 [#554](https://github.com/sktime/skbase/issues/554)).
 
 ### Use a tag to control whether pretrained parameters should be reset or not
+
 This would be equivalent to having the private class attribute
 `_PRETRAINED_ATTRIBUTES`. However, it introduces a tag for which the set of
 values is not clear beforehand. Tags store metadata, and they currently have a
@@ -135,6 +160,8 @@ expand the scope of tags.
 3. For the same circumstances above, check whether pretrained parameters are
    reset when `force_reset=True` is passed as an argument to `reset`,
    `clone`, and `set_params`.
-4. The reserved keyword argument `force_reset` should not be a `__init__` argument.
-5. Check if cloning and calling pretrain on the second instance does not change the pretrained parameters of the first instance.
-6. Check if setting incompatible hyperparameters with existing pretrained parameters raises an error.
+
+4. Check behavior of tests 2. when the model is used inside a pipeline or composition with a non-global model.
+5. The reserved keyword argument `force_reset` should not be a `__init__` argument.
+6. Check whether cloning and calling pretrain on the second instance does not change the pretrained parameters of the first instance.
+7. Check whether setting incompatible hyperparameters with existing pretrained parameters raises an error.
--- a/steps/27_pretrain_parameters/step.md
+++ b/steps/27_pretrain_parameters/step.md
@@ -2,7 +2,7 @@
 
 Contributors: @fkiraly @SimonBlanke @felipeangelimvieira
 
-This STEP has the idea of centralizing the discussion on edge-cases of global forecasters
+This STEP centralizes the discussion of edge cases for global forecasters
 and pretraining-compatible models with respect to pretrained parameters.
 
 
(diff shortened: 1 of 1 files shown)


steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+* users should be made aware when they are trying to set hyperparameters that are incompatible with the existing pretrained parameters, and an error should be raised in this case.
+  
+## Proposed solution
+
+### On API level
+
+1. Add a new argument to `reset`, `clone`, and `set_params` to control whether
+   pretrained parameters should be reset, with default value `False`, i.e.,
+   pretrained parameters are not reset by default. Suggested name:
+   `force_reset=False`.
```
SimonBlanke: Good idea in general, but I would be more specific for the parameter-name. `reset_pretrained` or `keep_pretrained` would be better in terms of semantics. If this is too specific we could also go with `reset_task`, since the model would switch from one task (for example during the fit) to the next.

SimonBlanke: hmm, but on this level (in skbase) we do not specifically look at pretraining. So a more general name might make sense after all.

felipeangelimvieira:
Regarding the shared memory issue: I think it is okay and good to share memory at first (for example, for raw pretrained weights available on hugging face). However, once there's a new pretraining (finetuning) step, we need to make sure that this new pretraining call does not change the weights inplace.

I've not included `pretrain_update` in this STEP. I see that `pretrain_update` might not be required for all forecasters and I would consider not having it in `BaseForecaster`, but it can be useful in a base class for Neural-network based global models or  as a mixin to avoid changing shared weights in-place and having side-effects.

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+pretraining state should be erased. Similarly, in a hyperparameter tuning
+pipeline where we want to find the best hyperparameters for pretraining, the
+pretrained parameters should be reset after each iteration. Therefore, we need a
+way to control whether pretrained parameters are reset, while keeping the
+default behavior of preserving non-task-specific parameters.
+
+### Problem 2: incompatibility of certain hyperparameters with existing pretrained parameters
+
+Consider the case where a user has a pretrained model with certain
+hyperparameters, such as `hidden_dim`, and sets them to new values. Because the
```
fkiraly, line 126: I do not understand this problem. What does it mean that a "pretrained model has certain hyperparameters".

I think it is crucial to start with a conceptual model at the very start that disambiguates and introduces clear terminology for what "parameters" we are talking about. I can think of at least five types of:

* *fitted* parameters set in `pretrain`
* *hyper*-parameters impacting only pretraining
* *fitted* parameters set in `fit`
* *hyper*-parameters impacting only the `fit`
* *hyper*-parameters impacting potentially both `pretrain` and `fit`

Whether this is what you mean or not, a clear and clean conceptual model is necessary to avoid confusion as well as "crooked designs".

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+1.  raise an error in this case
+2.  raise a warning and force the reset of pretrained parameters when
+    incompatible hyperparameters are set.
+
+Option 2 can silently change the behavior of the model without the user being
+fully aware of it, if they are not checking the logs. Therefore, we consider
+option 1 to be more explicit, because it makes the user aware of the exact
+behavior of the code while avoiding silent buggy behavior.
+
+This would require models to check whether a certain change in `__init__`
```
fkiraly, line 147: I think this paragraph already assumes a solution, so is an invalid conclusion at this stage.

You somehow assume that `__init__` happens *after* a call of `pretrain`? Or that there is a pretraining act that somehow happens before `__init__`? I do not think that makes sense in python, at least in the common paradigms, because at the start of anything you construct an instance of a class, and that, logically, must then happen before "pretraining".

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+shared memory between the original and cloned objects. This can lead to
+unintended side effects when one of the objects is modified, as it can affect
+the other object as well. Therefore, we need to ensure that when pretrained
+parameters are not reset during cloning, they are also properly copied to avoid
+shared memory issues.
+
+## Requirements
+
+* by default, `reset` should have the same behavior as `clone`
+* non-task-specific parameters should be cloned (or deep-copied if `.clone()` is not supported) and keep their values unchanged by default in `reset`, `clone`, and `set_params` operations
```
fkiraly: what is a "non-task-specific" parameter? I do not have that in my conceptual model... so, please explain.

fkiraly (changes requested):
Great STEP, although I am missing the conceptual model with examples. As seen above, I am not 100% sure what some terms mean.

May I hence ask for design 101, namely: conceptual model, with examples given for each non-obvious concept discussed? Optimally, three substantially distinct examples per non-obvious concept.

steps/27_pretrain_parameters/step.md
felipeangelimvieira: With the "pretrained fitted parameters", the blueprint of a model is not only `__init__` arguments but also the pretrained fitted parameters. The problem is `set_params` might try to change the corresponding `__init__` arguments in a way that they are incompatible with pretrained fitted parameters. In other words, if blueprint is (`__init__` kwargs, pretrained fitted parameters), we need to avoid incompatible combinations, such as pretrained weights that assume neural networks with `hidden_dim=128` and an `__init__` kwargs with `hidden_dim=129`.

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+* users should be able to reset non-task-specific parameters when needed
+* users should be made aware when they are trying to set hyperparameters that are incompatible with the existing pretrained parameters, and an error should be raised in this case.
+  
+## Proposed solution
+
+### On API level
+
+1. Add a new argument to `reset`, `clone`, and `set_params` to control whether
+   pretrained parameters should be reset, with default value `False`, i.e.,
+   pretrained parameters are not reset by default. Suggested name:
```
fkiraly, line 191: * adding this argument: agreed. Also, these should be exactly the same.
* why should the default be `False`? Please argue. This would be inconsistent with the "global" default via `scikit-base`.
* what about an alternative naming that makes semantic reference to the state diagram? E.g., `reset_to="pretrained"` or similar?

steps/27_pretrain_parameters/step.md
felipeangelimvieira: When I say:

> a certain change in `__init__` arguments

I refer to using `set_params`. Initialization always before `pretrain`, but users can `set_params` after `pretrain`.

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+   `force_reset=False`.
+2. Estimators should have a class attribute
+   `_PRETRAINED_ATTRIBUTES : list[str] | None` that lists the attributes that
```
fkiraly: * agree on high-level that there *should* be a centralized list of the parameters in question, this follows from multiple methods accessing it (however, it is not a "must"; since all methods that are not `reset` could internally call `reset` in theory)
* why a class attribute, and not a tag?
* why a class attribute/tag and not an object (dynamic) tag?

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+   belong to the model's pretraining phase and are not task-specific.
+3. `reset` and `clone` should both inspect that class attribute, reset the
```
fkiraly: with reset, you mean `del` / `delattr`, right? If so, better be explicit what "reset" means.

Normally, "reset" is read as "to a prior value" or "to a default value". In this case, the attributes would be removed via deletion though!

fkiraly: agree on high level

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+   listed attributes if `force_reset=True`, and keep them cloned/deep-copied if
+   `force_reset=False`. Every other attribute should be reset as usual.
+4. `set_params` should call a function `_check_set_params_compatibility(self, **params) -> None` that checks whether the new hyperparameters are compatible with the existing pretrained parameters, and raises an error if they are not. This function should be called before setting the new parameters. See the following skbase changes for more details on how to implement this.
```
fkiraly: disagree and I have a very bad feeling about this idea. I would advise not to follow through with it.

The reason is that we are not even doing anything similar with the "ordinary" hyperparameters, they could also be mutually incompatible.

This feels like something that should be left to individual models, could be done in `__post_init__` if the implementer desires.

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Skbase changes
```
fkiraly: I would strongly recommend to suggest an implementation pathway that has 2 stages:

* stage 1: changes in `sktime` only. This will allow us to move faster, find bugs and issues faster, user test faster, and not impact other users of `scikit-base` by accident.
* stage 2: move selected changes (e.g., plugin capability) to `scikit-base`.

I would actually be against going to stage 2 directly. I am 100% sure this would cause immense hassle jumping back/forth between the two code bases.

Luckily, everything should be easily overridable, handleable inside `sktime`.

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+
+For handling incompatible hyperparameters with existing pretrained parameters:
```
fkiraly: I disagree with this point and really need to see an example (see above)

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+* Option 1: Add a `set_params` plugin as in `clone`
+* Option 2: explicitly call a function `_check_set_params_compatibility(self, **params) -> None` in the base `set_params` implementation.
+
+For handling non-resettable attributes:
+
+* On the skbase side, we could add a `_NON_RESETTABLE_ATTRIBUTES_ATTR_NAME` class attribute for objects. This attribute specifies the name of the class attribute that lists non-resettable attributes.
```
fkiraly: Attribute or tag? See above.

Also, I find the naming `_NON_RESETTABLE_ATTRIBUTES_ATTR_NAME` a bit convoluted. Why not
`_ATTRIBUTE_DEFINING_ATTRS_THAT_ARE_NON_RESETTABLE_IDENTIFIER_STRING_ATTRIBUTE_NAME_REFERENCES_METADATA_RESOLUTION_LAYER_REFERENCE`

steps/27_pretrain_parameters/step.md
felipeangelimvieira: One example is weights in a Foundation Model. Non-task-specific parameters are parameters learned from data that can be used in new tasks. In that case, we might not want to erase such parameters when cloning() or resetting. Is this clear?

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+* The default is `None`, which maintains the current behavior of not having non-resettable attributes.
+* If set to a string, it should be the name of a class attribute that lists non-resettable attributes.
+* During reset and clone operations, it should check whether the attribute exists and raise an error if it is not set. It should then use the `force_reset` argument to decide whether to reset the attributes listed in that class attribute. Passing `force_reset` to objects that do not have non-resettable attributes should not cause any issues and should behave as usual.
+
+This would allow us to have a more general API for estimators that have attributes that are not task-specific.
```
fkiraly: can you argue why this should be in `scikit-base`?

The usual argument for introducing a layer of abstraction is that you have three (or more) distinct examples that you can capture with it.

* argument for `sktime` `BaseForecaster`: there are more than three such forecasters 👍 
* argument for `scikit-base`: `BaseForecaster`, ... ???

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+`reset` and would make it hard to control the compatibility of global models in
+pipelines and other compositions (see issue
+[#554](https://github.com/sktime/skbase/issues/554)).
+
+### Use a tag to control whether pretrained parameters should be reset or not
+
+This would be equivalent to having the private class attribute
+`_PRETRAINED_ATTRIBUTES`. However, it introduces a tag for which the set of
+values is not clear beforehand. Tags store metadata, and they currently have a
+finite set of possible values for every key. Adding a pretrained attributes tag
```
fkiraly: > Tags store metadata, and they currently have a
finite set of possible values for every key

Not true. There are plenty tags whose values are basically arbitrary alphanumeric strings.

The tag system, as used in `sktime` has evolved to be closer to `pyproject.toml` than its original "mostly boolean" origins.

Examples: `python_dependencies` (everything everywhere has this), `tests:skip_by_name`, `maintainers`, etc.

felipeangelimvieira: Good counterexamples, I did not consider them.

Then, what exactly defines `tag` and class attribute? Which decision rule you would have to use one or another?

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,167 @@
...
+would expand the scope of tags and make it less clear what values are expected
+for such a tag. On the other hand, having a private class attribute to list
+pretrained parameters is more explicit and easier to understand, and does not
+expand the scope of tags.
```
fkiraly: however, having a private class attribute complicates the extension contract. It looks "simpler", but it is not simpler if you think about how the extension template would look like!

I think design complexity can be really well estimated from "number of lines in a well written extension contract".

For a new tag, you just introduce two new lines - tag names and description what goes there.

For a new attribute, you have to add a whole new subsection with an explanation that refers to the attribute only (unless you are using private attributes as the default way to handle metadata)

steps/27_pretrain_parameters/step.md
fkiraly: I think I understand now, but I suspect I think this is the wrong conceptual mapping.

I believe you are referring to (again, without giving any examples so you force me to guess) to an example of, say, `FantasticFoundationModel(weights="pretrained_from_huggingface_location_here")`, so you get pretrained weights into, say, `self._weights = self.download_from(self.weights)` or similar.

Which means in this case you do not execute `pretrain` from `sktime`, the model has been pretrained.

Then you think, `self._weights` is special because it is not "task specific", and should not reset in `clone`, right?

steps/27_pretrain_parameters/step.md
fkiraly: if so, why do you conclude that "this would require models to check whether a certain *change in `__init__`*"

steps/27_pretrain_parameters/step.md
fkiraly: you are still not defining your terms - can you please do this and give examples in at least a stylized example of the different types of "parameters"?

steps/27_pretrain_parameters/step.md
felipeangelimvieira: > The reason is that we are not even doing anything similar with the "ordinary" hyperparameters, they could also be mutually incompatible.

This is usually handled after `super().__init__()` in forecasters. We check if the arguments are valid.

`__post_init__` is indeed a good alternative. I will add it to the STEP.

steps/27_pretrain_parameters/step.md
fkiraly: we introduced tags primarily so you can have:

* a programmatic way to query "what are all tags" - as opposed to a `dir` call or similar that returns all kinds of garbage
* a programmatic way to document them, by adding docstrings to tag registry classes
* instance level dynamic overrides (e.g., `capability:exogeneous` for pipelines),

felipeangelimvieira:
One problem I see at the moment (also appearing in different places) is that we need to handle the current state for every function call: 
* `__post_init__` will have a different behaviour when initializing from scratch vs pretrained. 
* In pretrain, we are also needing to have redirect to different behaviours if it is the first call or if the model has already been pretrained.

If the developer is not aware, not handling this properly can introduce bugs.

steps/27_pretrain_parameters/step.md
felipeangelimvieira: I was considering that scikit-base inherits the use-cases of sktime. I understand the idea of not propagating the complexity and changes to scikit-base or at least have a really strong reason to it.

fkiraly:
> * `__post_init__` will have a different behaviour when initializing from scratch vs pretrained. 

What does this mean? E.g., "from scratch vs pretrained"?

`__post_init__` currently is called exactly once, after `__init__`.

I do again think your semantic mapping is off.

What I would suggest: Models that have been *externally* pretrained do not count as "sktime pretrained".

Meaning, `FantasticFoundationModel(weights="pretrained_from_huggingface_location_here")` is initialized not in the `"pretrained"` state but in the `"new"` state. `"pretrained"` is only for models after at least one call of `sktime` `pretrain`.

This will hopefully solve the problems you see above (which arise because you try to force the state mapping to match your semantic perception rather than the API based definition).

steps/27_pretrain_parameters/step.md
fkiraly: ok, it looks like we agree 😁

felipeangelimvieira:
> __post_init__ currently is called exactly once, after __init__.

Sorry, for one moment I forgot that `set_params` acts in-place, changing the same object. You're right.
However, if using `__post_init__`, how to handle:

```python
model = FantasticFoundationModel(weights="pretrained_from_huggingface_location_here")
model.pretrain(X=X,y=y)
model.set_params(weights="incompatible_model_weights") # Should this trigger an error?
model.pretrain(X=X,y=y) # Model is already pretrained, but we have updated it with weights that are incompatible with its pretrained state!
```

fkiraly:
Currently:

* `set_params` calls `reset`
* `reset` calls `__init__`
* `__init__` calls `__post_init__`

So it should not require separate treatment of where the call comes from?

If you put the special parameter treatment in `reset`, most other methods will inherit it, though `clone` may require extra treatment.

steps/27_pretrain_parameters/step.md
fkiraly: yes - and also to avoid complexity blowup too early. Localizing the construction site when possible is a good idea...

steps/27_pretrain_parameters/step.md
SimonBlanke: If we don't do any changes in skbase we probably have to override the public reset of the BaseForecaster. We could avoid this by just implementing a hook into the skbase reset.

steps/27_pretrain_parameters/step.md
felipeangelimvieira: The phrase mentioned `__init__` arguments, not only `__init__`. In other words, I was refering to hyperparameters.

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,234 @@
...
+|`new`|`fit`|Yes|Yes|-|`fitted`|Load weights, save to pretrained attributes and bind to task|
+|`new`|`fit`|Yes|No|-|`fitted`|Call pretrain and then fit|
+|`pretrained` or `fitted`|`reset`|Yes|-|-|`pretrained`|Reset non-pretrained attributes, keep pretrained parameters|
+|`pretrained` or `fitted`|`reset(reset_pretrained=False)`|Yes|-|-|`pretrained`|Reset non-pretrained attributes, keep pretrained parameters|
+|`pretrained` or `fitted`|`reset(reset_pretrained=True)`|Yes|-|-|`new`|Reset pretrained attributes, reset all parameters|
+|`pretrained`|`set_params` on finetunable hyperparameters|Yes|-|Yes|`pretrained`|Call reset with `reset_pretrained=False` and change finetunable hyperparameters|
+|`pretrained`|`set_params` on non-finetunable hyperparameters|Yes|-|No|`new`|Change non-finetunable hyperparameters, reset all attributes and initialize|
+|`pretrained`|`pretrain` again|Yes|-|No|`pretrain`|Call reset(reset_pretrained=True) and then pretrain again, potentially going back to the same initial pretrained parameters|
+|`pretrained`|`pretrain` again|Yes|-|Yes|`pretrain`|Copy and incrementally change pretrained parameters, go to `pretrained` state|
```
felipeangelimvieira, line 185: @SimonBlanke  @fkiraly  I've added this table and I think we should find an agreement on this. Is there something missing? Would you add or change something in this table? We might want to map every state transition and its consequences here.

steps/27_pretrain_parameters/step.md
felipeangelimvieira: Added a section with definitions, please review

steps/27_pretrain_parameters/step.md
felipeangelimvieira: > why should the default be False? Please argue. This would be inconsistent with the "global" default via scikit-base.

For usage and compatibility with current compositions in sktime that call `reset` of inner estimators. As a practical example, consider calling `.reset()` in an `Ensemble` model that contains a pretrained estimator and a local estimator. The pretrained estimator should not go back to new state and lose its pretrained attributes. In most cases, I would say that users are willing to keep the pretrained state, and only reset `fit` attributes.

> what about an alternative naming that makes semantic reference to the state diagram? E.g., reset_to="pretrained" or similar?

I see the appeal of `reset_to`, but in this case I think the boolean argument may be preferable.

If `reset_to` is used, the valid values would essentially be new and pretrained. That is semantically equivalent to a boolean choice: either keep the pretrained parameters or do not. A boolean also avoids accepting arbitrary string values such as fitted, trained, or misspelled variants, unless we add extra validation logic.

So my preference would be to keep the boolean argument, with a name that makes the behavior explicit, e.g. keep_pretrained=True.

fkiraly:
>> why should the default be False? Please argue. This would be inconsistent with the "global" default via scikit-base.

> For usage and compatibility with current compositions in sktime that call `reset` of inner estimators. As a practical example, consider calling `.reset()` in an `Ensemble` model that contains a pretrained estimator and a local estimator. The pretrained estimator should not go back to new state and lose its pretrained attributes. In most cases, I would say that users are willing to keep the pretrained state, and only reset `fit` attributes.

One may then argue: should this not change the `reset` call instead, with a non-default `reset_pretrain=False`, to make `Ensemble` explicitly capable of `pretrain`?

For me, keeping `reset` default behaviour downwards compatible feels like an important point - just from the general point of ensuring downwards compatibility. If this were not the 1.0 release, we would actually need to keep `reset` downwards compatible! Which shows to me that we should perhaps be quite cautious with changing default behaviour.

> If `reset_to` is used, the valid values would essentially be new and pretrained. That is semantically equivalent to a boolean choice: either keep the pretrained parameters or do not. A boolean also avoids accepting arbitrary string values such as fitted, trained, or misspelled variants, unless we add extra validation logic.

My argument against boolean goes like this: I have seen too many cases in my life where initially we thought we had a dichotomous choice, only for shortly thereafter to realize that a third choice would also make sense. For this reason, I generally have a reluctance to add boolean flags unless it is really a clear yes/no question and not a "what/where" question.

More concisely, string-valued is much more upwards compatible than boolean, especially in cases where the variable in question is not clearly semantically boolean.

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,235 @@
...
+and pretraining-compatible models with respect to pretrained parameters.
+
+## Related issues
+* https://github.com/sktime/sktime/issues/10151
+* https://github.com/sktime/skbase/issues/554
+
+## Definitions
+
+### Global models
+Models that can be used with tasks different from those used to learn their parameters.
```
fkiraly, line 15: I think this definition does not hold since "task" is incorrectly used, when compared to current use in `sktime` - a task would be something like "forecasting", "time series classification".

But, both pretraining and fitting are typically for forecasting. Do you not mean: different data instances?

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Local models
+Models whose parameters and training are task-specific.
```
fkiraly, line 18: same, issue with use of "task".

steps/27_pretrain_parameters/step.md
```diff
@@ -0,0 +1,235 @@
...
+Global models whose weights and parameters are already stored somewhere and can be imported; there is no need for a learning phase on the user's machine to use them for new tasks.
+
+### Global non-foundation model (GNFM)
+Global models that need to be trained once before they can be used. There is no weight or parameter storage available. Example: ML models in a reduction framework.
+
+### Pretrained attributes
+These refer to attributes of the class changed or created during a `.pretrain` call.
+
+### Pretrained parameters
+These refer to the model parameters (e.g. neural network weights, an XGBoost fitted model for a reduction forecaster) that are global and not task-specific.
```
fkiraly, line 30: are weights parameters or attributes?

fkiraly (changes requested):
* thanks for adding the conceptual model. I think this is good, but has two problems:
    * us of the nebulous term "task". I think referring to the undefined "task" only shifts the problem of definition, it does not solve it. I think it is also used in a way that is not consistent with its current use in `sktime`.
    * Similarly, the "attributes" and "parameters" are not a good concept. "parameters" is not clearly defined, and it is not clear which of these would apply to typical neural network weights. Maybe an API based definition instead of an example based definition (e.g., using `get_params`, `get_fitted_params`) would be better here.
* I am still missing examples. I really, really, strongly recommend to be very explicit about running examples, e.g., `MyFavouriteFoundationModel` and spell out its parameters. Otherwise it will be hard to get aligned about what we are even talking about.

steps/27_pretrain_parameters/step.md
fkiraly: I think it is crucial to state what you mean by "reset". Do you mean "set to default value", or `delattr`? I think it should be the latter.

steps/27_pretrain_parameters/step.md
felipeangelimvieira: weights are parameters (in the mathematical sense) but are attributes in sktime objects.

steps/27_pretrain_parameters/step.md
felipeangelimvieira: Since there is no other "task" in sktime sense (such as classification, regression), there would be no ambiguity to define task as it is defined in this document (as detailed in the definition part) for this specific STEP. Data instances is also ok, I'll update it.

fkiraly:
@SimonBlanke has implemented a prototype in https://github.com/sktime/sktime/pull/10238 which departs from some proposals in this PR, but has not commented no this proposal resp the proposed differences, pros/cons.

@SimonBlanke, can I please ask you to comment on the design first, including different options, pros/cons, before you go ahead and implement? We need to avoid redundant work or duplicative work.

I am adding my comments here that I have also added in #10238, for convenience, and to keep discussion in one place.

First and foremost, before implementing, I think we *must* align on the exact function signatures and state change diagrams before starting to implement, or we will revert from a laborious implementation to the drawing board. Exploratory PRs are of course fine, but PRs without reference to this STEP or without engaging in discussion with the current design process are not helpful.

In particular, I would suggest discussing multiple options for the function signatures alongside the STEP, rather than implementing one specific option.

Some thoughts I have after looking at https://github.com/sktime/sktime/pull/10238:

* `set_params` has the signature `set_params(**kwargs)`, adding "special kwargs" for control of the state feels a crass departure from prior signature conventions.
* I would suggest to leave `set_params` as is, but change `reset` only to default to "keep pretrained parameters". If a user wants to reset all parameters, they can call `set_params` and `reset(to="new")` or similar
* I would suggest a signature `reset(self, to="pretrained")` as the signature, with values for `to` being exactly the potential values for `self._state`.
* should `clone` also have a target state? If `clone` should parallel `reset`, then the arguments of `clone` should probably also be the same as those of `reset`. However, if `clone` gets additional arguments, then the clone plugin manager needs to be extended to permit arguments.
* another option to manage the behaviour of `clone`, `reset`, `set_params` would be via config fields, i.e., `set_config`. I think there are serious pros and serious cons to consider for this approach.

steps/27_pretrain_parameters/step.md
SimonBlanke: If 'task' already has a certain meaning in sktime I would rather avoid the use here for the explanation of foundation models. 'forecasting problem instance' would be very descriptive.

steps/27_pretrain_parameters/step.md
SimonBlanke: > I think this paragraph already assumes a solution, so is an invalid conclusion at this stage.

I think Felipe and I discussed this solution in a meeting. Of course we are not 'set' on this.

SimonBlanke:
> set_params has the signature set_params(**kwargs), adding "special kwargs" for control of the state feels a crass departure from prior signature conventions.

I agree. This is a sloppy function signature. Not sure how it made it out of draft.

> I would suggest to leave set_params as is, but change reset only to default to "keep pretrained parameters". If a user wants to reset all parameters, they can call set_params and reset(to="new") or similar

I could not find a reason not to do it this way. I do not see an obstacle for the rest of the api unification.

> I would suggest a signature reset(self, to="pretrained") as the signature, with values for to being exactly the potential values for self._state.

This only works for only 2 of the 3 states. to="fitted" cannot be done (from what I can tell), so this should raise an error. 
We could also go with 'to_state' for the argument name.

> should clone also have a target state? If clone should parallel reset, then the arguments of clone should probably also be the same as those of reset. However, if clone gets additional arguments, then the clone plugin manager needs to be extended to permit arguments.

This makes sense in the long run, because of architectural consistency. I would move this to a later stage of development.

> another option to manage the behaviour of clone, reset, set_params would be via config fields, i.e., set_config. I think there are serious pros and serious cons to consider for this approach.

I see 'set_config' as more of a 'sticky' configuration that persists for the lifetime of the instance. It would make more longer pretrain, fit runs more complicated. A user that runs 'forecaster.reset()' would need to look elsewhere what the reset really does.

fkiraly:
I have thought about the design 


### Desiderata

* not breaking running code and core API contracts in general
* minimizing the change surface in `sktime` estimators
    * example: `ForecastingGridSearchCV` and pipelines employ `clone`
* API consistency with any other `scikit-base` estimator
    * example: recursive `set_params` calls where components are non-`sktime` estimators but `scikit-base` compatible
* extendability to all `scikit-base` estimators that have multiple states

### Use cases considered

The following use cases were considered in selecting the design

1. direct user calls to `set_params`, `clone`, `reset` of `sktime` forecasters
2. `reset` call in `BaseForecaster.fit`
3. `set_params` recursion to `set_params`, across potentially different object types
4. `ForecastingGridSearchCV` and other tuner calls to `set_params`
5. `ForecastingPipeline` and other composite estimator calls to `clone`
6. IMPORTANT: comparison of `clone` and `reset` call tree in composite estimators, noting that currently `reset` is not recursive, but `clone` is.
    * importantly, it should be noted that recursive `clone` is currently equivalent to a single `reset` call, since the `__init__` call guarantees to construct a "new" state object with the same attributes.
    * We should also consider the case where we have nesting as follows: `sktime` object - non-`sktime`-object, `sktime` object; or, forecaster - sth else without pretrain - forecaster/transformer, i.e., an intermediate nesting level without the new method(s)

### Proposed solution

* `set_params`, `reset` and `clone` remain as they are, for now
* in `BaseForecaster` (and later in `BaseTransformer`), we introduce private methods `_reset_at(state)`, `_clone_at(state)`, and `_set_params_at(state, params: dict)`, which default to `reset`, `clone` and `set_params` if the `capability:pretrain` tag is false.
    * there is a tag called `pretrain:fitted_params` or similar, which is a list of str containing all the parameters that the above methods should *not*  delete upon call.
    * `_set_params_at` is mostly a copy of `scikit-base` `set_params`, but calls `_reset_at` instead of `reset`, and passes on `state`.


### Rationale

#### Why new private methods?

* It is extremely dangerous to make changes to core API methods such as `reset`, `clone`, `set_params`, especially since external calls may assume their current API (delete everything), for instance in recursive `set_params` or `clone` calls.
* Current recursion logic in `set_params` and `clone` will break chains of passing the `state` parameter to composites (see case 3-6, and case 3, 6 especially), leading to unexpected behaviour. Component forecasters of composite forecasters may or may not be reset to pretrained state, if the recursion problem is not properly solved.
* none of the current designs floating around, e.g., this STEP of @felipeangelimvieira, or the experimental PR of @SimonBlanke indicate that the composition cases 3-6 were all clearly or explicitly considered, especially cases 3 and 6. This means there is, in my opinion, too high a risk to introduce breakage or unintended behaviour in API behaviour, even beyond forecasting.
    * working with new, forecaster-specific private methods helps localize the change. This can be a temporary measure until we understand the composition case, or how the public methods need to look like

#### Why `state` as new argument?

* `state` can be passed in recursions easily, if it is the same across all three methods and their calls - the methods make recursion calls, and also call each other
    * I do not have a strong opinion about the exact name
* it is closest to the conceptual model and semantics of what we want to do: we want to reset to the pretrained state.

#### Pros/cons

In terms of desiderata:

* not breaking running code and core API contracts in general
    * approach is conservative, and it does not change current API contracts.
    * the API does not change, instead code will need to change where currently direct calls to `clone` etc take place
* minimizing the change surface in `sktime` estimators
    * example: `ForecastingGridSearchCV` and pipelines employ `clone`
    * the proposed approach has a *higher* change surface than adding args to `clone` etc with defaults.
    * However, I think the benefits of derisking breaking API changes outweighs this disadvantage.
* API consistency with any other `scikit-base` estimator
    * example: recursive `set_params` calls where components are non-`sktime` estimators but `scikit-base` compatible
    * The above is the only alternative I can currently see which ensures API consistency at all!
    * Or, differently, currently, it feels very difficult to (a) come up with a way that maintains compatibility, that (b) does not change the entire base API, and which (c) the people likely to implement/coordinate (all very senior) can implement without significant breakage risk, e.g., for compositions across packages.
* extendability to all `scikit-base` estimators that have multiple states
    * Also a negative for this desideratum, the change remains localized to forecasters and/or transformers.
    * But, the localization is required to minimize the breakage risk for now.

### Alternatives considered

1. in `set_params`, remove `reset` call entirely. Handle `clone` and `reset` separately from `set_params`.
    * this would probably break use case 3, or require changes across every single object type which feels excessive (given that we are changing only forecasters and maybe transformations), and also impossible across the ecosystem.
2. Change signatures of `reset`, `clone`, `set_params`, (e.g., `reset_params=False` or similar parameters), but do not recurse them.
    * this will lead to unexpected behaviour in composites, see use cases 3-6.
    * however, if defaults are resetting to "pretrained" state, it would have the benefit that grid search and pipelines work "as expected" if they are "shallow", e.g., one component deep without code changes.
3. Change signatures of `reset`, `clone`, `set_params`, (e.g., `reset_params=False` or similar parameters), and add some form of state recursion to nested `clone` and `set_params` calls.
    * the main reason I am deciding against this is, I do not see how this would work, since calls can be to external packages (e.g., `scikit-learn`) which will not introduce the concept of an intermediate, pretrained state, and should also not need to since what we are doing is restricted to our own estimator types.
4. instead of the `pretrain:fitted_params` tag, use a class attribute
    * this would expand the extension contract beyond tags, it would be much more difficult in documentation (where would documentation go?). Developers expect documented tags, docs are built via the `registry._tags` module.

### Implementation notes

#### Test-driven implementation for composition use cases

It is crucial to set up test cases for use cases 3-6, if we implement this. That is, come up with explicit code examples (can be speculative) for these cases, and set them as tests.

Then I would proceed with test-driven development.

#### Simple/naive estimators

Having simple or naive estimators is always useful for testing. These should be written, so we do not have to use foundation models etc.

Examples for forecasters:

* forecast the average of historical data observed
* pick a random time series historically observed
* take an `sktime` distance, and pick a window in the time series historically observed with minimal distance to the observed data points. Forecast what follows that window.

Examples for transformations:

* multiplicatively normalize data across historical samples in `transform`, invert in `inverse_transform`
* global detrending - subtract average across historical samples in `transform`, add the average back in `inverse_transform`

#### Gradual implementation

This suggestion should be considered as part of a gradual plan:

1. implement in private methods, localized to `BaseForecaster` and `BaseTransformer`
2. once we think we have understood the challenges, we can make the methods public
3. potential in long-term: add state-aware handling of estimators to `scikit-base`. This will require a detailed design.

I would expect that there is at least *a year* between the different phases.

PR #52, opened 2026-07-27
[STEP] Recursive composite and native serialization

geetu040:
This pull request adds a design document for a unified serialization format across `sktime` and `skpro`. It combines recursive serialization of composite estimators with framework-native serialization for deep-learning and foundation models, while preserving the existing `save`/`load` API.

The proposal supersedes STEP 20 and specifies the archive layout, component and artifact handling, backward compatibility, implementation plan, and testing requirements.

Related discussions:

- https://github.com/sktime/skpro/issues/1072
- https://github.com/sktime/skpro/pull/1073
- https://github.com/sktime/sktime/issues/10450
- https://github.com/sktime/sktime/pull/10453
- https://github.com/sktime/sktime/issues/10582
- [HackMD: Serialization Architecture for sktime Deep Learning Estimators](https://hackmd.io/rShrR00UQrKcNo33P6dJBA)


fkiraly (changes requested):
Very very nice!

I mostly agree, one point I am unsure about is how to handle multiple applicable serialization formats. We should think about it - the current `save` allows, for instance, to pick pickle or cloudpickle.

fkiraly (approved):
ok, understood - the `serialization_format` is the baseline for metadata and base types, from which special serializers deviate. This also means though that nodes cannot pick their own serialization format. Perhaps something to keep in mind for a future design.

